PR #6, opened 2015-01-13
Fixes #8921 - Register to Apipie via plugin methods

mbacovsky:
The purpose of this is to be able to pre-build the Apipie docs during the package build and save some time during installation, when we have to build just the indexes.

Packaging part will be added to the https://github.com/theforeman/foreman-packaging/pull/426

### DIFF
--- a/lib/foreman_bootdisk/engine.rb
+++ b/lib/foreman_bootdisk/engine.rb
@@ -23,13 +23,12 @@ module ForemanBootdisk
     end
 
     initializer "foreman_bootdisk.apipie" do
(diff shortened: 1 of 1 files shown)


domcleal:
Ticket number is missing an '8'.

Does this require Foreman 1.8?  If so, please change the min version in engine.rb.

mbacovsky:
Ticket number was fixed. And you are right again, I thought `apipie_documented_controllers` is in 1.7 already. Fixed as well.

domcleal:
Thanks @mbacovsky!

PR #7, opened 2015-02-11
fixes #9252 - use new factories and explicitly use host's provision interface

domcleal:
@ares or @GregSutcliffe, is using the provision_interface explicitly here a good idea?  The job of this iPXE script is solely to find and configure the interface on which to retrieve the provision/kickstart template from Foreman.

http://ci.theforeman.org/job/test_plugin_pull_request/534/ running.

### DIFF
--- a/app/views/foreman_bootdisk/host.erb
+++ b/app/views/foreman_bootdisk/host.erb
@@ -5,11 +5,13 @@
 # Copy this template to customize it, the original is read-only.
 
 <%
-bootdisk_raise(N_('Host has no IP address defined')) if @host.ip.nil? || @host.ip.empty?
-bootdisk_raise(N_('Host has no subnet defined')) unless @host.subnet
-bootdisk_raise(N_('Host has no domain defined')) unless @host.domain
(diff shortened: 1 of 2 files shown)


domcleal:
Oh, the unattended seems to use the primary interface for identification, not the provision interface.  Is that a bug?

https://github.com/theforeman/foreman/blob/f5975f9/app/controllers/unattended_controller.rb#L144

ares:
1) using provision interface explicitly makes sense to me in this case

2) `find_host_by_ip_or_mac` is used on many places so I'm not entirely sure but from quick scan it looks to me as a bug that we don't use provision interface mac.

@GregSutcliffe do you agree?

GregSutcliffe:
@domcleal @ares I agree, this looks like a 1.7-style leftover. I suspect it hasn't come up because we're all using tokens by default now, but I can see how that causes an issue for bootdisk ;)

+1 for the bootdisk side too, this looks sane.

domcleal:
Thanks for the review and comments, I'll fix that in core.

domcleal:
https://github.com/theforeman/foreman/pull/2172 for the record.

PR #8, opened 2015-03-05
[WIP] Fixes #8146 - Add subnet-level generic image for template proxying

GregSutcliffe:
Posting this for approach-review before working on api/doc stuff

The goal here is to be able to have a copy of the "generic" image, but with the URL replaced with a proxy with the Template feature, so that the network isolation can work. This is clearly only going to work for subnets which can contact that particular proxy.

To achieve this, I've tricked `lib/foreman/renderer.rb` into using the template-proxying codepath by passing in a real Host in build mode. This feels like an acceptable requirement since we need to decide which subnet we're building for (additionally, all the bootdisks are built from the Host page, and all but the 'generic' image require build=true as well). The alternative is a bunch of extra scaffold to add a bootdisk button to the subnet or proxy page, which felt like overkill.

The only nasty bit is having to gsub out the token in the resulting template. Given that's how the core rendered currently works though, I'm not sure how to improve it.

Feedback welcome, and then I can update the api, tests, etc.

### DIFF
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -2,7 +2,7 @@ require 'uri'
 
 module ForemanBootdisk
   class DisksController < ::ApplicationController
-    before_filter :find_resource, :only => %w[host full_host]
+    before_filter :find_resource, :only => %w[host full_host subnet]
 
     def generic
       begin
@@ -37,6 +37,23 @@ module ForemanBootdisk
       host = @disk
       ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
         send_data File.read(iso), :filename => "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso"
+      end
+    end
+
+    def subnet
+      begin
+        subnet = @disk.try(:subnet)
+        fail unless subnet && subnet.try(:tftp)
+        tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      rescue => e
+        error _('Failed to render boot disk template: %s') % e
+        redirect_to :back
+        return
+      end
+
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+        name=subnet.try(:name)
+        send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
       end
     end
 
--- a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
+++ b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
@@ -11,12 +11,22 @@ module ForemanBootdisk::HostsHelperExt
         select_action_button(_('Boot disk'), {},
           display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'btn'),
+          display_bootdisk_for_subnet,
           display_bootdisk_link_if_authorized(_("Generic image"), {:controller => 'foreman_bootdisk/disks', :action => 'generic'}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Help"), {:controller => 'foreman_bootdisk/disks', :action => 'help'}, :class=>'btn')
         )
       )
     )
     host_title_actions_without_bootdisk(*args)
+  end
+
+  # need to wrap this one in a test for template proxy presence
+  def display_bootdisk_for_subnet
+    if (proxy = @host.try(:subnet).try(:tftp)) && proxy.features.map(&:name).include?('Templates')
+      display_bootdisk_link_if_authorized(_("Subnet '%s' image") % @host.subnet.name, {:controller => 'foreman_bootdisk/disks', :action => 'subnet', :id => @host}, :class=>'btn')
+    else
+      ""
+    end
   end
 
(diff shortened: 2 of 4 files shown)


GregSutcliffe:
Not quite sure what the merge conflicts are. I've pulled latest master and rebased against it...

domcleal:
Your commit parent is the 4.0.2 release commit.

GregSutcliffe:
Oh, my mistake. HEAD != HEAD:subnet when pushing.

domcleal:
How is the API going to work?  The API should be subnet-based, even if the UI isn't, so passing a host into the renderer service seems like a bad idea.

GregSutcliffe:
@domcleal I see your point. However, the renderer code in core requires a host with a valid token, in build mode, to use that code path. As it stands, I don't see how we could only use a subnet, other than by faking a host? I see you use a struct for the generic image, but I'm concerned about potentially leaving tokens lying around... any suggestions?

domcleal:
Yeah, that code isn't pleasant, we could do with improving it or making it more pluggable perhaps.

I think if you used a struct with a fixed token then it might be better for parsing out after, but it's still not a great solution.  It would at least allow you to make an API that isn't host-centric.

GregSutcliffe:
Agreed, the renderer needs to be more plugin-aware, somehow. I'll see what I can do.

GregSutcliffe:
@domcleal ok, so this seems better. The UI controller is determining the subnet from the host, but the API could be attached to the subnet directly. Both can then pass subnet into the generic renderer. Does that look ok to you?

lzap:
It looks like Greg is away for some time and I will be taking over this. @domcleal what is your recommendation here? Where should I start now? What are the biggest painpoints of this patch?

domcleal:
Yeah, that looks better internally, so the API can use a subnet as the base resource.

lzap:
I am starting with Foreman Bootdisk today, after reading README couple of questions:

* Does it use tokens for non-generic images? The README explicitly mentions MAC/IP.
* The file also mentions "finish" script, what can be "finish" script used for in this workflow (I assume since there is no image associated Foreman won't be able to ssh-connect).

domcleal:
Tokens aren't used for host images, MAC/IP ident is.  Tokens may be used for full host images.

Finish scripts are used for preseed provisioning.

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +40,23 @@ def full_host
       end
     end
 
+    def subnet
+      begin
+        subnet = @disk.try(:subnet)
```
lzap: Can someone explain me where does `@disk` come from? Grepping like crazy, no useful results :-)

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -19,6 +20,15 @@ def host_title_actions_with_bootdisk(*args)
     host_title_actions_without_bootdisk(*args)
   end
 
+  # need to wrap this one in a test for template proxy presence
+  def display_bootdisk_for_subnet
+    if (proxy = @host.try(:subnet).try(:tftp)) && proxy.features.map(&:name).include?('Templates')
+      display_bootdisk_link_if_authorized(_("Subnet '%s' image") % @host.subnet.name, {:controller => 'foreman_bootdisk/disks', :action => 'subnet', :id => @host}, :class=>'btn')
```
lzap: We need to add this permission to the role.

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +40,23 @@ def full_host
...
+        fail unless subnet && subnet.try(:tftp)
```
lzap: I think we should give clear error message here, otherwise we will only see the generic message ending up with `:`.

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +40,23 @@ def full_host
...
+        tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      rescue => e
+        error _('Failed to render boot disk template: %s') % e
+        redirect_to :back
+        return
+      end
+
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+        name=subnet.try(:name)
+        send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
```
lzap: Isn't `name` always defined?

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -5,14 +5,27 @@ class Renderer
...
+          Struct.new(:value).new('faketoken'),
+          subnet
+        )
+      else
+        @host = Struct.new(:token, :subnet).new(nil,nil)
+      end
+      rtmpl = unattended_render(tmpl.template)
+      # remove the token from subnet-level generation, since this is meant to be generic
+      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'') if @host.present?
+      rtmpl
```
lzap: Reminder: will need to test this with tokens disabled.

lzap:
Ok looks good, few comments. Please give me your opinions @GregSutcliffe or @domcleal and I will likely file a new PR based off this work as I expect @GregSutcliffe not to have free hands currently :)

app/controllers/foreman_bootdisk/disks_controller.rb
domcleal: Magic in the base controller, named after the base resource which is a "disk" in this case, except the class is actually a Host.

lzap: Holy cup of :coffee: 

Any comments for the ones below? Going to send the new PR today.

domcleal: Sorry, wasn't really sure what you wanted me to say, I assumed they were things @GregSutcliffe or you had to fix.

PR #9, opened 2015-04-03
#8146 - Subnet bootdisk support

lzap:
Based off https://github.com/theforeman/foreman_bootdisk/pull/8

Leaving Greg's patch as-is to keep credit and attaching another commit on top
of that. If you don't like that, I can squash keeping Greg's name.

TODO:

* test with tokens disabled/enabled
* cleanup error messages and failure scenarios
* add V2 API
* add tests

Anything else?

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,14 @@ TFTP settings are needed.
     <td>Yes</td>
     <td>No</td>
   </tr>
+  <tr>
+    <td>Subnet image</td>
+    <td>Yes</td>
+    <td>Yes</td>
+    <td>No</td>
+    <td>Yes</td>
+    <td>No</td>
+  </tr>
 </table>
 
 ### Per-host images
@@ -204,6 +212,33 @@ To generate from the command line on the Foreman server:
 Set `OUTPUT=/path` to change the output destination path (directory or file).
 It must be writable by the 'foreman' user.
 
+### Subnet images
+
+Subnet image is similar to generic image, but chain-loading is done via TFTP
+Smart Proxy assigned to Subnet of the host. The smart proxy must have
+"Templates" module enabled and configured.
+
+This image is generic for all hosts with provisioning NIC on that subnet.
+
+To generate the image from the web interface, view a host page, click the
+"Boot disk" button and select "Subnet image".
+
+To generate using the Hammer CLI, install the [hammer_cli_foreman_bootdisk](https://github.com/theforeman/hammer_cli_foreman_bootdisk)
+plugin and run:
+
+    hammer bootdisk subnet --subnet local.lan
+
+See the hammer_cli_foreman_bootdisk documentation for more advanced usage.
+
+To generate from the command line on the Foreman server:
+
+    foreman-rake bootdisk:generate:subnet NAME=subnet-name
+
+Note that foreman-rake does connect to the foreman-proxy.
+
+Set `OUTPUT=/path` to change the output destination path (directory or file).
+It must be writable by the 'foreman' user.
+
 ### Host group images
 
 TODO
@@ -243,6 +278,14 @@ the Foreman UI.
 * `bootdisk_mkiso_command` is the name of genisoimage/mkisofs on your OS
 * `bootdisk_cache_media` controls whether to cache OS boot files from
   installation media for full host images
+
+### RBAC
+
+This plugin provides "Boot disk access" role which requires the following
+permissions to be present to allow host and subnet image downloading:
+
+* `view_hosts`
+* `view_subnets`
 
 # Issues
 
--- a/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
@@ -11,6 +11,7 @@ module ForemanBootdisk
         end
 
         before_filter :find_resource, :only => :host
+
         skip_after_filter :log_response_body
 
         # no-op, but required for apipie documentation
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -8,7 +8,7 @@ module ForemanBootdisk
       begin
         tmpl = ForemanBootdisk::Renderer.new.generic_template_render
       rescue => e
-        error _('Failed to render boot disk template: %s') % e
+        error_rendering(e)
         redirect_to :back
         return
       end
@@ -23,7 +23,7 @@ module ForemanBootdisk
       begin
         tmpl = host.bootdisk_template_render
       rescue => e
-        error _('Failed to render boot disk template: %s') % e
+        error_rendering(e)
         redirect_to :back
         return
       end
@@ -41,19 +41,19 @@ module ForemanBootdisk
     end
 
     def subnet
+      host = @disk
       begin
-        subnet = @disk.try(:subnet)
-        fail unless subnet && subnet.try(:tftp)
+        subnet = host.try(:subnet) || raise(::Foreman::Exception.new(N_("Subnet is not assigned to the host %s"), host.name))
+        subnet.tftp || raise(::Foreman::Exception.new(N_("TFTP feature not enabled for subnet %s"), subnet.name))
         tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
       rescue => e
-        error _('Failed to render boot disk template: %s') % e
+        error_rendering(e)
         redirect_to :back
         return
       end
 
       ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
-        name=subnet.try(:name)
-        send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
+        send_data File.read(iso), :filename => "bootdisk_subnet_#{subnet.name}.iso"
       end
     end
 
@@ -64,6 +64,12 @@ module ForemanBootdisk
 
     def resource_scope(controller = controller_name)
       Host::Managed.authorized(:view_hosts)
+    end
+
+    def error_rendering(e)
+      msg = _('Failed to render boot disk template')
+      error("#{msg}: #{e.message}")
+      ::Foreman::Logging.exception(msg, e)
     end
   end
 end
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -2,7 +2,7 @@ require 'uri'
 
 module ForemanBootdisk
   class DisksController < ::ApplicationController
-    before_filter :find_resource, :only => %w[host full_host]
+    before_filter :find_resource, :only => %w[host full_host subnet]
 
     def generic
       begin
@@ -37,6 +37,23 @@ module ForemanBootdisk
       host = @disk
       ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
         send_data File.read(iso), :filename => "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso"
+      end
+    end
+
+    def subnet
+      begin
+        subnet = @disk.try(:subnet)
+        fail unless subnet && subnet.try(:tftp)
+        tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      rescue => e
+        error _('Failed to render boot disk template: %s') % e
+        redirect_to :back
+        return
+      end
+
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+        name=subnet.try(:name)
+        send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
       end
     end
 
--- a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
+++ b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
@@ -11,12 +11,22 @@ module ForemanBootdisk::HostsHelperExt
         select_action_button(_('Boot disk'), {},
           display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'btn'),
+          display_bootdisk_for_subnet,
           display_bootdisk_link_if_authorized(_("Generic image"), {:controller => 'foreman_bootdisk/disks', :action => 'generic'}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Help"), {:controller => 'foreman_bootdisk/disks', :action => 'help'}, :class=>'btn')
         )
       )
     )
     host_title_actions_without_bootdisk(*args)
+  end
+
+  # need to wrap this one in a test for template proxy presence
+  def display_bootdisk_for_subnet
+    if (proxy = @host.try(:subnet).try(:tftp)) && proxy.features.map(&:name).include?('Templates')
+      display_bootdisk_link_if_authorized(_("Subnet '%s' image") % @host.subnet.name, {:controller => 'foreman_bootdisk/disks', :action => 'subnet', :id => @host}, :class=>'btn')
+    else
+      ""
+    end
   end
 
   # Core Foreman helpers can't look up a URL against a mounted engine
--- a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
+++ b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
@@ -11,8 +11,10 @@ module ForemanBootdisk::HostsHelperExt
         select_action_button(_('Boot disk'), {},
           display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'btn'),
-          display_bootdisk_for_subnet,
+          content_tag(:li, "", :class => "divider"),
           display_bootdisk_link_if_authorized(_("Generic image"), {:controller => 'foreman_bootdisk/disks', :action => 'generic'}, :class=>'btn'),
+          display_bootdisk_for_subnet,
+          content_tag(:li, "", :class => "divider"),
           display_bootdisk_link_if_authorized(_("Help"), {:controller => 'foreman_bootdisk/disks', :action => 'help'}, :class=>'btn')
         )
       )
@@ -22,8 +24,8 @@ module ForemanBootdisk::HostsHelperExt
 
   # need to wrap this one in a test for template proxy presence
   def display_bootdisk_for_subnet
-    if (proxy = @host.try(:subnet).try(:tftp)) && proxy.features.map(&:name).include?('Templates')
-      display_bootdisk_link_if_authorized(_("Subnet '%s' image") % @host.subnet.name, {:controller => 'foreman_bootdisk/disks', :action => 'subnet', :id => @host}, :class=>'btn')
+    if (proxy = @host.try(:subnet).try(:tftp)) && proxy.has_feature?('Templates')
+      display_bootdisk_link_if_authorized(_("Subnet '%s' generic image") % @host.subnet.name, {:controller => 'foreman_bootdisk/disks', :action => 'subnet', :id => @host}, :class=>'btn')
     else
       ""
     end
--- a/app/models/concerns/foreman_bootdisk/host_ext.rb
+++ b/app/models/concerns/foreman_bootdisk/host_ext.rb
@@ -13,10 +13,7 @@ module ForemanBootdisk::HostExt
   end
 
   def bootdisk_chain_url(mac = self.mac, action = 'iPXE')
-    u = URI.parse(foreman_url(action))
-    u.query = "#{u.query}&mac=#{mac}"
-    u.fragment = nil
-    u.to_s
+    ForemanBootdisk::Renderer.format_bootdisk_chain_url(foreman_url(action), mac)
   end
 
   def bootdisk_raise(*args)
--- a/app/services/foreman_bootdisk/renderer.rb
+++ b/app/services/foreman_bootdisk/renderer.rb
@@ -7,8 +7,21 @@ module ForemanBootdisk
 
     def generic_template_render
       tmpl = ProvisioningTemplate.find_by_name(Setting[:bootdisk_generic_host_template]) || raise(::Foreman::Exception.new(N_('Unable to find template specified by %s setting'), 'bootdisk_generic_host_template'))
-      @host = Struct.new(:token, :subnet).new(nil, nil)
-      unattended_render(tmpl.template)
+
+      if subnet.present?
+        # rendering a subnet-level bootdisk requires tricking the renderer into thinking it has a
+        # valid host, with a token, and with a tftp proxy
+        @host = Struct.new(:token, :subnet).new(
+          Struct.new(:value).new('faketoken'),
+          subnet
+        )
+      else
+        @host = Struct.new(:token, :subnet).new(nil,nil)
+      end
+      rtmpl = unattended_render(tmpl.template)
+      # remove the token from subnet-level generation, since this is meant to be generic
+      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'') if @host.present?
+      rtmpl
     end
 
     def bootdisk_chain_url(action = 'iPXE')
--- a/app/services/foreman_bootdisk/renderer.rb
+++ b/app/services/foreman_bootdisk/renderer.rb
@@ -4,8 +4,9 @@ module ForemanBootdisk
   class Renderer
     include ::Foreman::Renderer
     include Rails.application.routes.url_helpers
+    delegate :logger, :to => :ForemanBootdisk
 
-    def generic_template_render
+    def generic_template_render(subnet = nil)
       tmpl = ProvisioningTemplate.find_by_name(Setting[:bootdisk_generic_host_template]) || raise(::Foreman::Exception.new(N_('Unable to find template specified by %s setting'), 'bootdisk_generic_host_template'))
 
       if subnet.present?
@@ -18,15 +19,20 @@ module ForemanBootdisk
       else
         @host = Struct.new(:token, :subnet).new(nil,nil)
       end
-      rtmpl = unattended_render(tmpl.template)
+      modified_template = unattended_render(tmpl.template)
       # remove the token from subnet-level generation, since this is meant to be generic
-      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'') if @host.present?
-      rtmpl
+      modified_template.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'')
+      modified_template
     end
 
-    def bootdisk_chain_url(action = 'iPXE')
-      u = URI.parse(foreman_url(action))
-      u.query = "#{u.query}&mac="
+    def bootdisk_chain_url(mac = '', action = 'iPXE')
+      Renderer.format_bootdisk_chain_url(foreman_url(action), mac)
+    end
+
+    def self.format_bootdisk_chain_url(url, mac)
+      u = URI.parse(url)
+      ForemanBootdisk.logger.warn("Foreman or proxy is configured with HTTPS, probably not supported by iPXE: #{u}") if u.scheme == 'https'
+      u.query = "#{u.query}&mac=#{mac}"
       u.fragment = nil
       u.to_s
     end
--- a/lib/foreman_bootdisk/engine.rb
+++ b/lib/foreman_bootdisk/engine.rb
@@ -31,7 +31,7 @@ module ForemanBootdisk
         requires_foreman '>= 1.9'
 
         security_block :bootdisk do |map|
-          permission :download_bootdisk, {:'foreman_bootdisk/disks' => [:generic, :host, :full_host, :help],
+          permission :download_bootdisk, {:'foreman_bootdisk/disks' => [:generic, :host, :full_host, :subnet, :help],
                                           :'foreman_bootdisk/api/v2/disks' => [:generic, :host]}
         end
 
@@ -57,5 +57,9 @@ module ForemanBootdisk
         puts "#{ForemanBootdisk::ENGINE_NAME}: skipping engine hook (#{e.to_s})"
       end
     end
+  end
+
+  def self.logger
+    Foreman::Logging.logger('foreman_bootdisk')
   end
 end
--- a/lib/tasks/bootdisk.rake
+++ b/lib/tasks/bootdisk.rake
@@ -45,6 +45,18 @@ namespace :bootdisk do
         puts "Wrote #{output}"
       end
     end
+
+    desc 'Generate a subnet disk for a specific subnet. NAME=subnet, OUTPUT=path'
+    task :subnet => :environment do
+      subnet = Subnet.find_by_name(ENV['NAME']) || raise("cannot find subnet '#{ENV['NAME']}', specify NAME=subnet")
+      subnet.tftp || raise(::Foreman::Exception.new(N_("TFTP feature not enabled for subnet %s"), subnet.name))
+      tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl, :dir => outputdir) do |image|
+        output = ENV['OUTPUT'] || File.join(outputdir, "bootdisk_subnet_#{subnet.name}.iso")
+        FileUtils.cp image, output
+        puts "Wrote #{output}"
+      end
+    end
   end
 end
 
--- a/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
+++ b/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
@@ -4,7 +4,7 @@ class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
   setup :setup_bootdisk
 
   test "should generate generic image" do
-    ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/generic host template/))).yields("temp ISO")
+    ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/disk generic host template/))).yields("temp ISO")
     File.expects(:read).with("temp ISO").returns("ISO image")
     get :generic, {}
     assert_response :success
@@ -12,10 +12,12 @@ class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
   end
 
   describe "#host" do
+    setup :setup_org_loc
+    setup :setup_subnet
     setup :setup_host
 
     test "should generate host image" do
-      ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/host template/))).yields("temp ISO")
+      ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/disk host template/))).yields("temp ISO")
       File.expects(:read).with("temp ISO").returns("ISO image")
       get :host, {:id => @host.name}
       assert_response :success
--- a/test/functional/foreman_bootdisk/disks_controller_test.rb
+++ b/test/functional/foreman_bootdisk/disks_controller_test.rb
@@ -4,7 +4,7 @@ class ForemanBootdisk::DisksControllerTest < ActionController::TestCase
   setup :setup_bootdisk
 
   test "should generate generic image" do
-    ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/generic host template/))).yields("temp ISO")
+    ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/disk generic host template/))).yields("temp ISO")
     File.expects(:read).with("temp ISO").returns("ISO image")
     get :generic, {}, set_session_user
     assert_response :success
@@ -12,10 +12,12 @@ class ForemanBootdisk::DisksControllerTest < ActionController::TestCase
   end
 
   describe "#host" do
+    setup :setup_org_loc
+    setup :setup_subnet
     setup :setup_host
 
     test "should generate host image" do
-      ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/host template/))).yields("temp ISO")
+      ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/disk host template/))).yields("temp ISO")
       File.expects(:read).with("temp ISO").returns("ISO image")
       get :host, {:id => @host.name}, set_session_user
       assert_response :success
@@ -28,6 +30,28 @@ class ForemanBootdisk::DisksControllerTest < ActionController::TestCase
       get :full_host, {:id => @host.name}, set_session_user
       assert_response :success
       assert_equal "ISO image", @response.body
+    end
+
+    test "should generate subnet image" do
+      ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/disk generic host template/))).yields("temp ISO")
+      ForemanBootdisk::Renderer.any_instance.stubs(:bootdisk_chain_url).yields("http://smart-proxy.lan")
+      File.expects(:read).with("temp ISO").returns("ISO image")
+      get :subnet, {:id => @host.name}, set_session_user
+      assert_empty flash[:error]
+      assert_response :success
+      assert_equal "ISO image", @response.body
+    end
+  end
+
+  describe "#host without tftp" do
+    setup :setup_org_loc
+    setup :setup_subnet_no_tftp
+    setup :setup_host
+
+    test "should not generate subnet image" do
+      get :subnet, {:id => @host.name}, set_session_user
+      assert_match(/Failed.*: TFTP feature not enabled/, flash[:error])
+      assert_response :redirect
     end
   end
 
--- a/test/test_plugin_helper.rb
+++ b/test/test_plugin_helper.rb
@@ -19,10 +19,22 @@ class ActionController::TestCase
     load File.join(File.dirname(__FILE__), '..', 'db', 'seeds.d', '50-bootdisk_templates.rb')
   end
 
-  def setup_host
+  def setup_org_loc
     disable_orchestration
-    org, loc = FactoryGirl.create(:organization), FactoryGirl.create(:location)
-    subnet = FactoryGirl.create(:subnet, :gateway => '10.0.1.254', :dns_primary => '8.8.8.8', :organizations => [org], :locations => [loc])
-    @host = FactoryGirl.create(:host, :managed, :subnet => subnet, :ip => subnet.network.sub(/0$/, '4'), :organization => org, :location => loc)
+    request.env["HTTP_REFERER"] = "/history"
+    @org, @loc = FactoryGirl.create(:organization), FactoryGirl.create(:location)
   end
+
+  def setup_subnet
+    @subnet = FactoryGirl.create(:subnet, :tftp, :gateway => '10.0.1.254', :dns_primary => '8.8.8.8', :organizations => [@org], :locations => [@loc])
+  end
+
+  def setup_subnet_no_tftp
+    @subnet = FactoryGirl.create(:subnet, :gateway => '10.0.1.254', :dns_primary => '8.8.8.8', :organizations => [@org], :locations => [@loc])
+  end
+
+  def setup_host
+    @host = FactoryGirl.create(:host, :managed, :subnet => @subnet, :ip => @subnet.network.sub(/0$/, '4'), :organization => @org, :location => @loc)
+  end
+
 end
--- a/app/controllers/foreman_bootdisk/api/v2/subnet_disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/api/v2/subnet_disks_controller.rb
@@ -1,0 +1,37 @@
+require 'uri'
+
+module ForemanBootdisk::Api::V2
+  class SubnetDisksController < ::Api::V2::BaseController
+    include ::Api::Version2
+
+    resource_description do
+      api_base_url "/bootdisk/api"
+    end
+
+    before_filter :find_resource, :only => :subnet
+
+    skip_after_filter :log_response_body
+
+    # no-op, but required for apipie documentation
+    api :GET, '', N_('Subnet boot disks')
+    def index; end
+
+    api :GET, '/subnets/:subnet_id', N_('Download subnet generic image')
+    param :subnet_id, :identifier_dottable, :required => true
+    def subnet
+      subnet = @subnet_disk
+      subnet.tftp || raise(::Foreman::Exception.new(N_("TFTP feature not enabled for subnet %s"), subnet.name))
+      tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+        name=subnet.name
+        send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
+      end
+    end
+
+    private
+
+    def resource_scope
+      Subnet.authorized('view_subnets')
+    end
+  end
+end
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -23,7 +23,7 @@ class ForemanBootdisk::ISOGenerator
         iso_suffix = iso9660_filename(suffix)
         iso_f0 = iso9660_filename(f[0].to_s) + '_' + iso_suffix
         tmpl.gsub!(f[0].to_s + '-' + suffix, iso_f0)
-        Rails.logger.debug("Boot file #{iso_f0}, source #{f[1]}")
+        ForemanBootdisk.logger.debug("Boot file #{iso_f0}, source #{f[1]}")
         [iso_f0, f[1]]
       end
     end
@@ -107,10 +107,10 @@ class ForemanBootdisk::ISOGenerator
       file.binmode
 
       if use_cache && !(contents = Rails.cache.fetch(uri, :raw => true)).nil?
-        Rails.logger.info("Retrieved #{uri} from local cache (use foreman-rake tmp:cache:clear to empty)")
+        ForemanBootdisk.logger.info("Retrieved #{uri} from local cache (use foreman-rake tmp:cache:clear to empty)")
         file.write(contents)
       else
-        Rails.logger.info("Fetching #{uri}")
+        ForemanBootdisk.logger.info("Fetching #{uri}")
         write_cache = use_cache
         uri = URI(uri)
         Net::HTTP.start(uri.host, uri.port) do |http|
@@ -126,7 +126,7 @@ class ForemanBootdisk::ISOGenerator
     end
 
     if write_cache
-      Rails.logger.debug("Caching contents of #{uri}")
+      ForemanBootdisk.logger.debug("Caching contents of #{uri}")
       Rails.cache.write(uri, File.read(path), :raw => true)
     end
   end
--- a/app/views/foreman_bootdisk/disks/help.html.erb
+++ b/app/views/foreman_bootdisk/disks/help.html.erb
@@ -29,4 +29,12 @@
   <p>
     <%= _('The OS install continues using the installation media configured in Foreman, and it will typically configure static networking, depending on how the OS iPXE template is configured.') %>
   </p>
+
+  <h2><%= _('Subnet image') %></h2>
+  <p>
+    <%= _('Subnet image is similar to generic image, but chain-loading is done via TFTP Smart Proxy assigned to Subnet of the host. The smart proxy must have "Templates" module enabled and configured.') %>
+  </p>
+  <p>
+    <%= _('This image is generic for all hosts with provisioning NIC on that subnet.') %>
+  </p>
 </div>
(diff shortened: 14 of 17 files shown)


lzap:
@daobrien fyi in this patch I will incorporate some documentation improvements we've talked about, stay CCed

lzap:
While testing this, I found that template smart-proxy plugin must not have `https` enabled, otherwise iPXE fails with error http://ipxe.org/err/3c0920 (unsupported protocol). I will add an error to the build phase if this is found to be https.

lzap:
@domcleal do you happen to know why I see this response on my proxy 20 times when I generate the subnet ISO once?

```
127.0.0.1 - - [27/May/2015 17:14:48] "GET /unattended/templateServer HTTP/1.1" 200 54 0.0005
lzapx.brq.redhat.com - - [27/May/2015:17:14:48 CEST] "GET /unattended/templateServer HTTP/1.1" 200 54
- -> /unattended/templateServer
```

domcleal:
No idea, I don't know this code.  Use a debugger?

lzap:
TODO note: create a wiki page for ERF error "'Proxy is configured with https, not supported for iPXE'"

lzap:
Finally got it working with https://github.com/theforeman/smart-proxy/pull/279 ! Now I can clenaup this patch and finalize this. Man that was tough to understand how this magic work...

lzap:
Ok this is ready for review, except API V2.

lzap:
Oh and documetnation! Of course.

lzap:
Pushed documentation as well.

lzap:
Finished with rake implementation, what's left is API/CLI. @domcleal mind taking a look next week? Thanks!

lzap:
This is now ready for full review.

README.md
```diff
@@ -203,6 +211,34 @@ To generate from the command line on the Foreman server:
 Set `OUTPUT=/path` to change the output destination path (directory or file).
 It must be writable by the 'foreman' user.
 
+### Subnet images
+
+Subnet image is similar to generic image, but chain-loading is done via TFTP
+Smart Proxy assigned to Subnet of the host. The smart proxy must have
+"Templates" plugin installed, enabled and configured to report http URL (https
```
domcleal: "module enabled", not "plugin installed"

README.md
```diff
@@ -203,6 +211,34 @@ To generate from the command line on the Foreman server:
...
+is not supported by iPXE at the moment).
```
domcleal: depends on the iPXE build, no?

README.md
```diff
@@ -203,6 +211,34 @@ To generate from the command line on the Foreman server:
...
+
+This image is generic for all hosts from the given subnet.
+
+To generate the image from the web interface, view a host page, click the
+"Boot disk" button and select "Subnet image".
+
+To generate using the Hammer CLI, install the [hammer_cli_foreman_bootdisk](https://github.com/theforeman/hammer_cli_foreman_bootdisk)
+plugin and run:
+
+    hammer bootdisk subnet
```
domcleal: https://github.com/theforeman/hammer_cli_foreman_bootdisk/pull/4 suggests it's a flag under host?

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -2,13 +2,14 @@
...
   class DisksController < ::ApplicationController
-    before_filter :find_resource, :only => %w[host full_host]
+    before_filter :find_resource, :only => %w[host full_host subnet]
 
     def generic
       begin
         tmpl = ForemanBootdisk::Renderer.new.generic_template_render
       rescue => e
         error _('Failed to render boot disk template: %s') % e
+        logger.debug e.backtrace.join("\n")
```
domcleal: use `Foreman::Logging.exception`

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -24,6 +25,7 @@ def host
         tmpl = host.bootdisk_template_render
       rescue => e
         error _('Failed to render boot disk template: %s') % e
+        logger.debug e.backtrace.join("\n")
```
domcleal: ditto

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +42,25 @@ def full_host
       end
     end
 
+    def subnet
+      host = @disk
+      begin
+        subnet = host.try(:subnet) || raise("Subnet is not assigned to the host #{host.name}")
```
domcleal: use error or Foreman::Exception please

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +42,25 @@ def full_host
...
+        subnet.try(:tftp) || raise("TFTP feature not enabled on subnet #{subnet.name}")
```
domcleal: ditto

domcleal: Also, don't use .try if subnet is always present - it's checked above.

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +42,25 @@ def full_host
...
+        tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      rescue => e
+        error _('Failed to render boot disk template: %s') % e
+        logger.debug e.backtrace.join("\n")
```
domcleal: ditto, Foreman::Logging

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +42,25 @@ def full_host
...
+        redirect_to :back
+        return
+      end
+
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+        name=subnet.try(:name)
```
domcleal: spacing around `=`

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -11,6 +11,7 @@ def host_title_actions_with_bootdisk(*args)
         select_action_button(_('Boot disk'), {},
           display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'btn'),
+          display_bootdisk_for_subnet,
```
domcleal: This is OK, but should it not (also?) be under the Subnets index?

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,15 +4,30 @@ module ForemanBootdisk
   class Renderer
     include ::Foreman::Renderer
     include Rails.application.routes.url_helpers
+    delegate :logger, :to => :Rails
```
domcleal: use the plugin logger

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,15 +4,30 @@ module ForemanBootdisk
...
+        @host = Struct.new(:token, :subnet).new(
+          Struct.new(:value).new('faketoken'),
+          subnet
+        )
+      else
+        @host = Struct.new(:token, :subnet).new(nil,nil)
+      end
+      rtmpl = unattended_render(tmpl.template)
+      # remove the token from subnet-level generation, since this is meant to be generic
+      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'') if @host.present?
```
domcleal: When will `@host` be nil?

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -27,13 +27,23 @@ def generic
 
         api :GET, '/hosts/:host_id', N_('Download host image')
         param :full, :bool, :required => false, :desc => N_('True for full, false for basic reusable image')
+        param :subnet, :bool, :required => false, :desc => N_('Generate subnet image (not compatible with "full" option)')
         param :host_id, :identifier_dottable, :required => true
         def host
+          raise ("Full and Subnet parameters are exclusive") if params[:full] && params[:subnet]
```
domcleal: don't just `raise`, use `render_message` or similar?

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -27,13 +27,23 @@ def generic
...
           host = @disk
           if params[:full]
             ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
               send_data File.read(iso), :filename => "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso"
             end
+          elsif params[:subnet]
+            subnet = host.try(:subnet) || raise("Subnet is not assigned to the host #{host.name}")
```
domcleal: less raise, more proper error reporting

app/controllers/foreman_bootdisk/disks_controller.rb
domcleal: String also needs extracting.

domcleal:
It seems the API and rake task are rather host-centric, it should be based around a subnet instead as this isn't specific to a host.

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
lzap: Yeah, can drop it there as well.

app/services/foreman_bootdisk/renderer.rb
lzap: Creating plugin logger and initializing it in the engine.rb as a class attribute.

app/services/foreman_bootdisk/renderer.rb
lzap: In the Parallel Universe, of course :-)

lzap:
Thanks, do you have an idea how to add a button to Subnets? For hosts there is this separate `hosts_title_actions` method, but subnets do use the generic `title_actions`.

lzap:
I figured out how to add something in the index title bar. The same comment (host-centric) applies to UI. I will create a POC patch and refactor ticket to move subnet bootdisk from hosts UI page to subnets UI page. But for the initial version, I would like to keep the UI as is, if you don't mind.

https://github.com/theforeman/foreman/pull/2461

lzap:
Ok done with all your comments, thanks!

```diff
diff --git a/README.md b/README.md
index 2ab2182..910a3a7 100644
--- a/README.md
+++ b/README.md
@@ -216,8 +216,7 @@ It must be writable by the 'foreman' user.
 
 Subnet image is similar to generic image, but chain-loading is done via TFTP
 Smart Proxy assigned to Subnet of the host. The smart proxy must have
-"Templates" plugin installed, enabled and configured to report http URL (https
-is not supported by iPXE at the moment).
+"Templates" module installed, enabled and configured.
 
 This image is generic for all hosts from the given subnet.
 
@@ -227,13 +226,13 @@ To generate the image from the web interface, view a host page, click the
 To generate using the Hammer CLI, install the [hammer_cli_foreman_bootdisk](https://github.com/theforeman/hammer_cli_foreman_bootdisk)
 plugin and run:
 
-    hammer bootdisk subnet
+    hammer bootdisk subnet --subnet local.lan
 
 See the hammer_cli_foreman_bootdisk documentation for more advanced usage.
 
 To generate from the command line on the Foreman server:
 
-    foreman-rake bootdisk:generate:subnet NAME=foo.example.com
+    foreman-rake bootdisk:generate:subnet NAME=subnet-name
 
 Note that foreman-rake does connect to the foreman-proxy.
 
diff --git a/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb b/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
index 6b0e83c..524586f 100644
--- a/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
@@ -27,23 +27,13 @@ module ForemanBootdisk
 
         api :GET, '/hosts/:host_id', N_('Download host image')
         param :full, :bool, :required => false, :desc => N_('True for full, false for basic reusable image')
-        param :subnet, :bool, :required => false, :desc => N_('Generate subnet image (not compatible with "full" option)')
         param :host_id, :identifier_dottable, :required => true
         def host
-          raise ("Full and Subnet parameters are exclusive") if params[:full] && params[:subnet]
           host = @disk
           if params[:full]
             ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
               send_data File.read(iso), :filename => "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso"
             end
-          elsif params[:subnet]
-            subnet = host.try(:subnet) || raise("Subnet is not assigned to the host #{host.name}")
-            subnet.try(:tftp) || raise("TFTP feature not enabled on subnet #{subnet.name}")
-            tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
-            ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
-              name=subnet.try(:name)
-              send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
-            end
           else
             tmpl = host.bootdisk_template_render
             ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
@@ -52,6 +42,19 @@ module ForemanBootdisk
           end
         end
 
+        api :GET, '/subnets/:subnet_id', N_('Download subnet image')
+        param :subnet_id, :identifier_dottable, :required => true
+        def subnet
+          id = params[:id]
+          subnet = Subnet.find(id) || render_message(_("Cannot find subnet with id %s") % id) && return
+          subnet.try(:tftp) || render_message(_("TFTP feature not enabled on subnet %s") % subnet.name) && return
+          tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+          ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+            name=subnet.try(:name)
+            send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
+          end
+        end
+
         private
 
         def resource_scope
diff --git a/app/controllers/foreman_bootdisk/disks_controller.rb b/app/controllers/foreman_bootdisk/disks_controller.rb
index 6736c88..3aa5866 100644
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -8,8 +8,7 @@ module ForemanBootdisk
       begin
         tmpl = ForemanBootdisk::Renderer.new.generic_template_render
       rescue => e
-        error _('Failed to render boot disk template: %s') % e
-        logger.debug e.backtrace.join("\n")
+        ::Foreman::Logging.exception(_('Failed to render boot disk template'), e)
         redirect_to :back
         return
       end
@@ -24,8 +23,7 @@ module ForemanBootdisk
       begin
         tmpl = host.bootdisk_template_render
       rescue => e
-        error _('Failed to render boot disk template: %s') % e
-        logger.debug e.backtrace.join("\n")
+        ::Foreman::Logging.exception(_('Failed to render boot disk template'), e)
         redirect_to :back
         return
       end
@@ -46,17 +44,16 @@ module ForemanBootdisk
       host = @disk
       begin
         subnet = host.try(:subnet) || raise("Subnet is not assigned to the host #{host.name}")
-        subnet.try(:tftp) || raise("TFTP feature not enabled on subnet #{subnet.name}")
+        subnet.tftp || raise("TFTP feature not enabled on subnet #{subnet.name}")
         tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
       rescue => e
-        error _('Failed to render boot disk template: %s') % e
-        logger.debug e.backtrace.join("\n")
+        ::Foreman::Logging.exception(_('Failed to render boot disk template'), e)
         redirect_to :back
         return
       end
 
       ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
-        name=subnet.try(:name)
+        name = subnet.name
         send_data File.read(iso), :filename => "bootdisk_subnet_#{name}.iso"
       end
     end
diff --git a/app/services/foreman_bootdisk/iso_generator.rb b/app/services/foreman_bootdisk/iso_generator.rb
index 016a8c1..b7c044a 100644
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -23,7 +23,7 @@ class ForemanBootdisk::ISOGenerator
         iso_suffix = iso9660_filename(suffix)
         iso_f0 = iso9660_filename(f[0].to_s) + '_' + iso_suffix
         tmpl.gsub!(f[0].to_s + '-' + suffix, iso_f0)
-        Rails.logger.debug("Boot file #{iso_f0}, source #{f[1]}")
+        ForemanBootdisk.logger.debug("Boot file #{iso_f0}, source #{f[1]}")
         [iso_f0, f[1]]
       end
     end
@@ -107,10 +107,10 @@ class ForemanBootdisk::ISOGenerator
       file.binmode
 
       if use_cache && !(contents = Rails.cache.fetch(uri, :raw => true)).nil?
-        Rails.logger.info("Retrieved #{uri} from local cache (use foreman-rake tmp:cache:clear to empty)")
+        ForemanBootdisk.logger.info("Retrieved #{uri} from local cache (use foreman-rake tmp:cache:clear to empty)")
         file.write(contents)
       else
-        Rails.logger.info("Fetching #{uri}")
+        ForemanBootdisk.logger.info("Fetching #{uri}")
         write_cache = use_cache
         uri = URI(uri)
         Net::HTTP.start(uri.host, uri.port) do |http|
@@ -126,7 +126,7 @@ class ForemanBootdisk::ISOGenerator
     end
 
     if write_cache
-      Rails.logger.debug("Caching contents of #{uri}")
+      ForemanBootdisk.logger.debug("Caching contents of #{uri}")
       Rails.cache.write(uri, File.read(path), :raw => true)
     end
   end
diff --git a/app/services/foreman_bootdisk/renderer.rb b/app/services/foreman_bootdisk/renderer.rb
index a6208fd..577ae4f 100644
--- a/app/services/foreman_bootdisk/renderer.rb
+++ b/app/services/foreman_bootdisk/renderer.rb
@@ -4,7 +4,7 @@ module ForemanBootdisk
   class Renderer
     include ::Foreman::Renderer
     include Rails.application.routes.url_helpers
-    delegate :logger, :to => :Rails
+    delegate :logger, :to => :ForemanBootdisk
 
     def generic_template_render(subnet=nil)
       tmpl = ProvisioningTemplate.find_by_name(Setting[:bootdisk_generic_host_template]) || raise(::Foreman::Exception.new(N_('Unable to find template specified by %s setting'), 'bootdisk_generic_host_template'))
@@ -21,7 +21,7 @@ module ForemanBootdisk
       end
       rtmpl = unattended_render(tmpl.template)
       # remove the token from subnet-level generation, since this is meant to be generic
-      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'') if @host.present?
+      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'')
       rtmpl
     end
 
diff --git a/config/routes.rb b/config/routes.rb
index 4384e9d..1f904cd 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,6 +14,7 @@ ForemanBootdisk::Engine.routes.draw do
       get 'generic', :to => 'disks#generic'
       constraints(:id => /[^\/]+/) do
         get 'hosts/:id', :to => 'disks#host'
+        get 'subnets/:id', :to => 'disks#subnet'
       end
     end
   end
diff --git a/lib/foreman_bootdisk/engine.rb b/lib/foreman_bootdisk/engine.rb
index 7412aae..3364a36 100644
--- a/lib/foreman_bootdisk/engine.rb
+++ b/lib/foreman_bootdisk/engine.rb
@@ -58,4 +58,8 @@ module ForemanBootdisk
       end
     end
   end
+
+  def self.logger
+    Foreman::Logging.logger('foreman_bootdisk')
+  end
 end
diff --git a/lib/tasks/bootdisk.rake b/lib/tasks/bootdisk.rake
index c9ebb95..0eb7f1e 100644
--- a/lib/tasks/bootdisk.rake
+++ b/lib/tasks/bootdisk.rake
@@ -46,20 +46,18 @@ namespace :bootdisk do
       end
     end
 
-    desc 'Generate a subnet disk for a specific host/subnet. NAME=fqdn, OUTPUT path'
+    desc 'Generate a subnet disk for a specific subnet. NAME=subnet, OUTPUT=path'
     task :subnet => :environment do
-      host = Host::Base.find_by_name(ENV['NAME']) || raise("cannot find host '#{ENV['NAME']}', specify NAME=fqdn")
-      subnet = host.try(:subnet) || raise("Subnet is not assigned to the host #{host.name}")
+      subnet = Subnet.find_by_name(ENV['NAME']) || raise("cannot find subnet '#{ENV['NAME']}', specify NAME=subnet")
       subnet.try(:tftp) || raise("TFTP feature not enabled on subnet #{subnet.name}")
       tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
       ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl, :dir => outputdir) do |image|
         name=subnet.try(:name)
-        output = ENV['OUTPUT'] || File.join(outputdir, "bootdisk_subnet_#{name}")
+        output = ENV['OUTPUT'] || File.join(outputdir, "bootdisk_subnet_#{name}.iso")
         FileUtils.cp image, output
         puts "Wrote #{output}"
       end
     end
-
   end
 end
 
diff --git a/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb b/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
index 79ebe93..ae72c8d 100644
--- a/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
+++ b/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
@@ -34,7 +34,7 @@ class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
       ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/host template/))).yields("temp ISO")
       ForemanBootdisk::Renderer.any_instance.stubs(:bootdisk_chain_url).yields("http://smart-proxy.lan")
       File.expects(:read).with("temp ISO").returns("ISO image")
-      get :host, {:id => @host.name, :subnet => true}
+      get :subnet, {:id => @host.subnet.id}
       assert_empty flash[:error]
       assert_response :success
       assert_equal "ISO image", @response.body
```

lzap:
For some reason I started getting this: https://github.com/freerange/mocha/issues/199

Let's see if we see the same on Jenkins...

lzap:
Fixed tests, this is now ready to review.

lzap:
Dafaq. Tests get aborted. [test]

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,15 +4,30 @@ module ForemanBootdisk
...
+        @host = Struct.new(:token, :subnet).new(
+          Struct.new(:value).new('faketoken'),
+          subnet
+        )
+      else
+        @host = Struct.new(:token, :subnet).new(nil,nil)
+      end
+      rtmpl = unattended_render(tmpl.template)
+      # remove the token from subnet-level generation, since this is meant to be generic
+      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'')
```
lzap: I am going to remove this `gsub!` because I am unable to generate host-based generic image with a token. I simply don't see them here, therefore this is useless?

lzap: Oh now I see, "faketoken" is being inserted above! Hm.

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -19,6 +20,15 @@ def host_title_actions_with_bootdisk(*args)
     host_title_actions_without_bootdisk(*args)
   end
 
+  # need to wrap this one in a test for template proxy presence
+  def display_bootdisk_for_subnet
+    if (proxy = @host.try(:subnet).try(:tftp)) && proxy.features.map(&:name).include?('Templates')
```
ohadlevy: can we extract it to a method? it repeats a few times and we might not want to relay on tftp feature..

lzap: I don't think it is worth it, we actually do not use it in this form anywhere else. But improving this to:

        if (proxy = @host.try(:subnet).try(:tftp)) && proxy.has_feature?('Templates')

lzap:
Ok I've verified and everything works now just fine:

* subnet image
* host image
* full host image

When templating is enabled. Unless there are other concerns, this is RTM.

README.md
```diff
@@ -204,6 +212,33 @@ To generate from the command line on the Foreman server:
 Set `OUTPUT=/path` to change the output destination path (directory or file).
 It must be writable by the 'foreman' user.
 
+### Subnet images
+
+Subnet image is similar to generic image, but chain-loading is done via TFTP
+Smart Proxy assigned to Subnet of the host. The smart proxy must have
+"Templates" module installed, enabled and configured.
```
domcleal: installed?

README.md
```diff
@@ -204,6 +212,33 @@ To generate from the command line on the Foreman server:
...
+
+This image is generic for all hosts from the given subnet.
```
domcleal: with provisioning NICs on that subnet?

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -42,6 +42,19 @@ def host
           end
         end
 
+        api :GET, '/subnets/:subnet_id', N_('Download subnet image')
+        param :subnet_id, :identifier_dottable, :required => true
+        def subnet
+          id = params[:id]
+          subnet = Subnet.find(id) || render_message(_("Cannot find subnet with id %s") % id) && return
```
domcleal: This controller's scoped to managed hosts, perhaps put it into another controller and ensure it's scoped correctly?

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -42,6 +42,19 @@ def host
...
+          subnet.try(:tftp) || render_message(_("TFTP feature not enabled on subnet %s") % subnet.name) && return
```
domcleal: why `try`?

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -42,6 +42,19 @@ def host
...
+          tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+          ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+            name=subnet.try(:name)
```
domcleal: ditto, why `try`

app/controllers/foreman_bootdisk/disks_controller.rb
domcleal: Still needs fixing.

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +40,24 @@ def full_host
       end
     end
 
+    def subnet
+      host = @disk
+      begin
+        subnet = host.try(:subnet) || raise("Subnet is not assigned to the host #{host.name}")
+        subnet.tftp || raise("TFTP feature not enabled on subnet #{subnet.name}")
```
domcleal: ditto

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -40,6 +40,24 @@ def full_host
...
+        tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      rescue => e
+        error_rendering(e)
+        redirect_to :back
+        return
+      end
+
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
+        name = subnet.name
```
domcleal: Not really needed?

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
domcleal: Needs doing?

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
domcleal: Needs doing?

app/models/concerns/foreman_bootdisk/host_ext.rb
```diff
@@ -14,6 +14,7 @@ def bootdisk_template_render
 
   def bootdisk_chain_url(mac = self.mac, action = 'iPXE')
     u = URI.parse(foreman_url(action))
+    raise(::Foreman::Exception.new(N_('Foreman or proxy is configured with https, not supported for iPXE'))) if u.scheme == 'https'
```
domcleal: `HTTPS`

domcleal: Isn't this very particular to the build of iPXE?  It seems a little limiting to me if it was rebuilt with the appropriate CA.

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,15 +4,30 @@ module ForemanBootdisk
...
+      rtmpl
     end
 
     def bootdisk_chain_url(action = 'iPXE')
       u = URI.parse(foreman_url(action))
+      raise(::Foreman::Exception.new(N_('Foreman or proxy is configured with https, not supported for iPXE'))) if u.scheme == 'https'
```
domcleal: ditto `HTTPS` comment

lib/tasks/bootdisk.rake
```diff
@@ -45,6 +45,19 @@ namespace :bootdisk do
         puts "Wrote #{output}"
       end
     end
+
+    desc 'Generate a subnet disk for a specific subnet. NAME=subnet, OUTPUT=path'
+    task :subnet => :environment do
+      subnet = Subnet.find_by_name(ENV['NAME']) || raise("cannot find subnet '#{ENV['NAME']}', specify NAME=subnet")
+      subnet.try(:tftp) || raise("TFTP feature not enabled on subnet #{subnet.name}")
```
domcleal: Why try?  Also Foreman::Exception as per the UI controller.

lib/tasks/bootdisk.rake
```diff
@@ -45,6 +45,19 @@ namespace :bootdisk do
...
+      tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+      ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl, :dir => outputdir) do |image|
+        name=subnet.try(:name)
```
domcleal: ditto try, useless variable

test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
```diff
@@ -29,5 +29,14 @@ class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
       assert_response :success
       assert_equal "ISO image", @response.body
     end
+
+    test "should generate subnet host image" do
```
domcleal: subnet generic?

test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
```diff
@@ -29,5 +29,14 @@ class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
...
+      ForemanBootdisk::ISOGenerator.expects(:generate).with(has_entry(:ipxe => regexp_matches(/host template/))).yields("temp ISO")
```
domcleal: Why's this matching host template, is it using the host-specific rather than generic template?  I'm not sure that I follow.

test/functional/foreman_bootdisk/disks_controller_test.rb
```diff
@@ -29,6 +29,27 @@ class ForemanBootdisk::DisksControllerTest < ActionController::TestCase
...
+      assert_response :success
+      assert_equal "ISO image", @response.body
+    end
+  end
+
+  describe "#host without tftp" do
+    setup :setup_host_no_tftp
+
+    test "should not generate subnet image" do
+      @host.subnet.tftp = nil
```
domcleal: What's the helper above for, if you set it to nil again here?

test/test_plugin_helper.rb
```diff
@@ -21,6 +21,15 @@ def setup_templates
 
   def setup_host
     disable_orchestration
+    request.env["HTTP_REFERER"] = "/history"
+    org, loc = FactoryGirl.create(:organization), FactoryGirl.create(:location)
+    subnet = FactoryGirl.create(:subnet, :tftp, :gateway => '10.0.1.254', :dns_primary => '8.8.8.8', :organizations => [org], :locations => [loc])
+    @host = FactoryGirl.create(:host, :managed, :subnet => subnet, :ip => subnet.network.sub(/0$/, '4'), :organization => org, :location => loc)
+  end
+
+  def setup_host_no_tftp
```
domcleal: Can you remove the duplication?

README.md
lzap: > installed?

Haven't realized we ship it by default.

-- 
Later,
 Lukas #lzap Zapletal

README.md
lzap: Fixed.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
lzap: I will just remove the scope, it's `only => :host` so there is no reason to break this apart.

lzap: Ok done, I've also found that we require `view_hosts` permission, but it is missing from the `Bootdisk access` role, so adding it as well as `view_subnets` there. I will add a migration to add these to existing role perhaps, opinion?

app/controllers/foreman_bootdisk/disks_controller.rb
lzap: Yeah all fixed, copy and paste error I believe. Can't explain...

app/controllers/foreman_bootdisk/disks_controller.rb
lzap: Hmm I see this one fixed already. Did not refresh on your side?

app/models/concerns/foreman_bootdisk/host_ext.rb
lzap: Hmm I will throw a warning in the server log then instead.

app/services/foreman_bootdisk/renderer.rb
lzap: Btw I am under impression that this method is no logner needed. Need to test this, but I see no reason why the one from host extension cannot be used.

lzap: Ok not true, at least I was able to deduplicate code and introduce `def self.format_bootdisk_chain_url(url, mac)` in the renderer.

test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
lzap: Yeah as well as in apipie description, fixed.

lib/tasks/bootdisk.rake
lzap: Fixed both.

test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
lzap: Fixed, the problem with these tests is that "host template" matches "generic host template", so we don't actually test those :-) Will fix the regexps.

lzap: Fyi this was just a typo that was not catched by tests.

test/functional/foreman_bootdisk/disks_controller_test.rb
lzap: Leftover.

test/test_plugin_helper.rb
lzap: Sure, all comments fixed! Amending.

lib/foreman_bootdisk/engine.rb
```diff
@@ -31,11 +31,11 @@ class Engine < ::Rails::Engine
         requires_foreman '>= 1.9'
 
         security_block :bootdisk do |map|
-          permission :download_bootdisk, {:'foreman_bootdisk/disks' => [:generic, :host, :full_host, :help],
+          permission :download_bootdisk, {:'foreman_bootdisk/disks' => [:generic, :host, :full_host, :subnet, :help],
                                           :'foreman_bootdisk/api/v2/disks' => [:generic, :host]}
         end
 
-        role "Boot disk access", [:download_bootdisk]
+        role "Boot disk access", [:download_bootdisk, :view_hosts, :view_subnets]
```
lzap: I don't know why this fails:
    /var/lib/workspace/workspace/test_plugin_pull_request@2/database/postgresql/ruby/1.9.3/foreman/app/models/role.rb:131:in `add_permissions': some permissions were not found (ArgumentError)

The new permissions are added via core seeds... strange.

lzap: Ok this is chicken-egg problem. I think we need to move roles from engine to seeds and adopt the foreman-core approach.

lzap: I have removed the `role` statement and moved towards seed.d/ approach. I am going to deprecate the `role` DSL and modify plugin documentation once this is confirmed.

lzap:
@ohadlevy amended last additional change - reordered the menu and added dividers as well

ohadlevy:
@lzap @domcleal anything missing?

lzap:
We are all set:

```
$ hammer bootdisk generic && hammer bootdisk host --host bdsktest.local.lan && hammer bootdisk host --host bdsktest.local.lan --full && hammer bootdisk subnet --subnet local.lan
Successfully downloaded generic disk image to bootdisk_192.168.122.1.iso
Successfully downloaded host disk image to bdsktest.local.lan.iso
Successfully downloaded host disk image to bdsktest.local.lan_20150722_0848.iso
Successfully downloaded subnet disk image to bootdisk_subnet_local.lan.iso
```

lzap:
@domcleal from the tag, I assume this needs to be tested, right?

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,16 +4,35 @@ module ForemanBootdisk
   class Renderer
     include ::Foreman::Renderer
     include Rails.application.routes.url_helpers
+    delegate :logger, :to => :ForemanBootdisk
 
-    def generic_template_render
+    def generic_template_render(subnet=nil)
```
shlomizadok: nitpick: space between param and value

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,16 +4,35 @@ module ForemanBootdisk
...
+        # rendering a subnet-level bootdisk requires tricking the renderer into thinking it has a
+        # valid host, with a token, and with a tftp proxy
+        @host = Struct.new(:token, :subnet).new(
+          Struct.new(:value).new('faketoken'),
+          subnet
+        )
+      else
+        @host = Struct.new(:token, :subnet).new(nil,nil)
+      end
+      rtmpl = unattended_render(tmpl.template)
```
shlomizadok: A meaningful name?

shlomizadok:
Tested, works as expected
Two minor nitpicks

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -10,7 +10,6 @@ class DisksController < ::Api::V2::BaseController
           api_base_url "/bootdisk/api"
         end
 
-        before_filter :find_resource, :only => :host
```
domcleal: I'd prefer if the new feature was in its own subnet-specific controller so as not to remove the use of helpers.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -42,10 +41,17 @@ def host
...
 
-        private
-
-        def resource_scope
-          Host::Managed.authorized('view_hosts')
+        api :GET, '/subnets/:subnet_id', N_('Download subnet generic image')
+        param :subnet_id, :identifier_dottable, :required => true
+        def subnet
+          id = params[:id]
+          subnet = Subnet.authorized('view_subnets').find(id) || render_message(_("Cannot find subnet with id %s") % id) && return
```
domcleal: capitalise "ID"

app/controllers/foreman_bootdisk/disks_controller.rb
domcleal: Still not fixed here.

lib/foreman_bootdisk/engine.rb
```diff
@@ -31,12 +31,10 @@ class Engine < ::Rails::Engine
         requires_foreman '>= 1.9'
 
         security_block :bootdisk do |map|
-          permission :download_bootdisk, {:'foreman_bootdisk/disks' => [:generic, :host, :full_host, :help],
+          permission :download_bootdisk, {:'foreman_bootdisk/disks' => [:generic, :host, :full_host, :subnet, :help],
                                           :'foreman_bootdisk/api/v2/disks' => [:generic, :host]}
         end
 
-        role "Boot disk access", [:download_bootdisk]
```
domcleal: Can you restore this please?  I don't know why you were putting view_hosts and view_subnets permissions in it anyway, so I don't see the need to use a DB seed.

domcleal:
Help page also needs updating with this new type.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
domcleal: No, definitely don't do that, the plugin shouldn't grant access to view these resources.

README.md
domcleal: Well, it's part of smart-proxy.  It's just a matter of configuration.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
lzap: I agree in general, but this particular helper is actually misleading, we search for host in disks controller. The code is cleaner without the helper actually in my opinion. In addition, we will be moving subnet image button to subnets page, therefore the actions will likely move somewhere else.

domcleal: Yeah, so it would be better to split hosts and subnet related resources into different controllers so we *can* use the helpers and we're more likely to pick up fixes and features from Foreman.

This doesn't have anything to do with the UI, as far as I can tell.

lib/foreman_bootdisk/engine.rb
lzap: This was my attempt to provide saner default role for new users. Bootdisk role by itself is unusable as you cannot generate host-images and subnet-images without this permission (either unfiltered or with a filter).

Seed script was to workaround chicken and egg problem, but I found a different solution in https://github.com/theforeman/foreman/pull/2572.

Anyway I am removing this from this patch and leaving this as is.

domcleal: > This was my attempt to provide saner default role for new users. Bootdisk role by itself is unusable as you cannot generate host-images and subnet-images without this permission (either unfiltered or with a filter).

That's very much intentional as this plugin doesn't know what permissions a user is meant to have, so it works additively.

lzap:
@domcleal @shlomizadok amended all your comments, removed seed script and put back the DSL role statement:

```diff
diff --git a/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb b/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
index 1a6baf5..80d6ca1 100644
--- a/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
@@ -45,7 +45,7 @@ module ForemanBootdisk
         param :subnet_id, :identifier_dottable, :required => true
         def subnet
           id = params[:id]
-          subnet = Subnet.authorized('view_subnets').find(id) || render_message(_("Cannot find subnet with id %s") % id) && return
+          subnet = Subnet.authorized('view_subnets').find(id) || render_message(_("Cannot find subnet with ID %s") % id) && return
           subnet.tftp || raise(::Foreman::Exception.new(N_("TFTP feature not enabled for subnet %s"), subnet.name))
           tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
           ForemanBootdisk::ISOGenerator.generate(:ipxe => tmpl) do |iso|
diff --git a/app/services/foreman_bootdisk/renderer.rb b/app/services/foreman_bootdisk/renderer.rb
index cfe6e67..5da68ca 100644
--- a/app/services/foreman_bootdisk/renderer.rb
+++ b/app/services/foreman_bootdisk/renderer.rb
@@ -6,7 +6,7 @@ module ForemanBootdisk
     include Rails.application.routes.url_helpers
     delegate :logger, :to => :ForemanBootdisk
 
-    def generic_template_render(subnet=nil)
+    def generic_template_render(subnet = nil)
       tmpl = ProvisioningTemplate.find_by_name(Setting[:bootdisk_generic_host_template]) || raise(::Foreman::Exception.new(N_('Unable to find template specified by %s setting'), 'bootdisk_generic_host_template'))
 
       if subnet.present?
@@ -19,10 +19,10 @@ module ForemanBootdisk
       else
         @host = Struct.new(:token, :subnet).new(nil,nil)
       end
-      rtmpl = unattended_render(tmpl.template)
+      modified_template = unattended_render(tmpl.template)
       # remove the token from subnet-level generation, since this is meant to be generic
-      rtmpl.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'')
-      rtmpl
+      modified_template.gsub!(/(?<=iPXE\?)token=faketoken\&(?=mac=)/,'')
+      modified_template
     end
 
     def bootdisk_chain_url(mac = '', action = 'iPXE')
diff --git a/app/views/foreman_bootdisk/disks/help.html.erb b/app/views/foreman_bootdisk/disks/help.html.erb
index 4c7ca6e..cf6f771 100644
--- a/app/views/foreman_bootdisk/disks/help.html.erb
+++ b/app/views/foreman_bootdisk/disks/help.html.erb
@@ -29,4 +29,12 @@
   <p>
     <%= _('The OS install continues using the installation media configured in Foreman, and it will typically configure static networking, depending on how the OS iPXE template is configured.') %>
   </p>
+
+  <h2><%= _('Subnet image') %></h2>
+  <p>
+    <%= _('Subnet image is similar to generic image, but chain-loading is done via TFTP Smart Proxy assigned to Subnet of the host. The smart proxy must have "Templates" module enabled and configured.') %>
+  </p>
+  <p>
+    <%= _('This image is generic for all hosts with provisioning NIC on that subnet.') %>
+  </p>
 </div>
diff --git a/lib/foreman_bootdisk/engine.rb b/lib/foreman_bootdisk/engine.rb
index 5e5d1ca..3364a36 100644
--- a/lib/foreman_bootdisk/engine.rb
+++ b/lib/foreman_bootdisk/engine.rb
@@ -35,6 +35,8 @@ module ForemanBootdisk
                                           :'foreman_bootdisk/api/v2/disks' => [:generic, :host]}
         end
 
+        role "Boot disk access", [:download_bootdisk]
+
         allowed_template_helpers :bootdisk_chain_url, :bootdisk_raise
         apipie_documented_controllers ["#{ForemanBootdisk::Engine.root}/app/controllers/foreman_bootdisk/api/v2/*.rb"]
       end
```

lzap:
Added a README note about the role and permissions:

```markdown
### RBAC

This plugin provides "Boot disk access" role which requires the following
permissions to be present to allow host and subnet image downloading:

* `view_hosts`
* `view_subnets`
```

app/controllers/foreman_bootdisk/disks_controller.rb
lzap: Ok finally corrected.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
lzap: To summarise, you want to move the subnet action from v2/disks_controller into new controller like v2/subnet_disks_controller?

I don't follow what this change has anything to do with "pikcing up fixes and features from Foreman"

lzap: Ok, since this would require breaking existing route ('/hosts/:host_id') I will do that in a separate pull request. This kind of changes should really be done in a clean refactoring separate commit. Changes needs to be done in CLI as well.

lzap:
@shlomizadok can you please retest? couple of minor edits are summarised in the diff above. Thanks!

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
domcleal: > To summarise, you want to move the subnet action from v2/disks_controller into new controller like v2/subnet_disks_controller?

Correct.



> Ok, since this would require breaking existing route ('/hosts/:host_id') I will do that in a separate pull request.

No need to change the existing actions or controller at the moment, they work as-is and could be renamed later.

shlomizadok:
Tested with the recent changes. Works as expected.

lzap:
Ok finally got to this, moved subnet into separate v2 controller, updated CLI as well. @domcleal @shlomizadok @ohadlevy @GregSutcliffe

lzap:
@shlomizadok one last test run please?

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -48,5 +65,11 @@ def help
     def resource_scope(controller = controller_name)
       Host::Managed.authorized(:view_hosts)
     end
+
+    def error_rendering(e)
+      msg = _('Failed to render boot disk template')
+      error(msg + ': ' + e.message)
```
shlomizadok: `error("#{msg}: #{e.message})` ?

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -11,7 +11,10 @@ def host_title_actions_with_bootdisk(*args)
         select_action_button(_('Boot disk'), {},
           display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'btn'),
+          content_tag(:li, "", :class=>"divider"),
```
shlomizadok: nitpick: spacing between class & divider

lib/tasks/bootdisk.rake
```diff
@@ -45,6 +45,18 @@ namespace :bootdisk do
         puts "Wrote #{output}"
       end
     end
+
+    desc 'Generate a subnet disk for a specific subnet. NAME=subnet, OUTPUT=path'
+    task :subnet => :environment do
+      subnet = Subnet.find_by_name(ENV['NAME']) || raise("cannot find subnet '#{ENV['NAME']}', specify NAME=subnet")
```
shlomizadok: raise unless?

shlomizadok:
@lzap - tested. Works well :+1: 
Had one nitpick with the raising an error in the rake task

lib/tasks/bootdisk.rake
lzap: This rake task follows shell-like syntax, its in the whole file so I just align with this style.

The other ones fixed, pushed.

lzap:
```diff
diff --git a/app/controllers/foreman_bootdisk/disks_controller.rb b/app/controllers/foreman_bootdisk/disks_controller.rb
index d182dd0..d481eab 100644
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -68,7 +68,7 @@ module ForemanBootdisk
 
     def error_rendering(e)
       msg = _('Failed to render boot disk template')
-      error(msg + ': ' + e.message)
+      error("#{msg}: #{e.message}")
       ::Foreman::Logging.exception(msg, e)
     end
   end
diff --git a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
index 8bb2e68..56f9b77 100644
--- a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
+++ b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
@@ -11,10 +11,10 @@ module ForemanBootdisk::HostsHelperExt
         select_action_button(_('Boot disk'), {},
           display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'btn'),
           display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'btn'),
-          content_tag(:li, "", :class=>"divider"),
+          content_tag(:li, "", :class => "divider"),
           display_bootdisk_link_if_authorized(_("Generic image"), {:controller => 'foreman_bootdisk/disks', :action => 'generic'}, :class=>'btn'),
           display_bootdisk_for_subnet,
-          content_tag(:li, "", :class=>"divider"),
+          content_tag(:li, "", :class => "divider"),
           display_bootdisk_link_if_authorized(_("Help"), {:controller => 'foreman_bootdisk/disks', :action => 'help'}, :class=>'btn')
         )
       )
```

ohadlevy:
@shlomizadok can you verify your nit is resolved? otherwise +1 to merge.

shlomizadok:
:+1: lgtm

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -11,6 +11,7 @@ class DisksController < ::Api::V2::BaseController
         end
 
         before_filter :find_resource, :only => :host
+
```
ohadlevy: nitpick: extra whitespace

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,16 +4,35 @@ module ForemanBootdisk
...
-    def generic_template_render
+    def generic_template_render(subnet = nil)
       tmpl = ProvisioningTemplate.find_by_name(Setting[:bootdisk_generic_host_template]) || raise(::Foreman::Exception.new(N_('Unable to find template specified by %s setting'), 'bootdisk_generic_host_template'))
-      @host = Struct.new(:token, :subnet).new(nil, nil)
-      unattended_render(tmpl.template)
+
+      if subnet.present?
+        # rendering a subnet-level bootdisk requires tricking the renderer into thinking it has a
+        # valid host, with a token, and with a tftp proxy
+        @host = Struct.new(:token, :subnet).new(
```
ohadlevy: i wasn't sure as it was not explicit in the new tests, but do you cover this usage case with a test?

lzap: This is indeed covered with tests.

domcleal:
Merged as 367b1e6, d344850, thanks @lzap.

PR #11, opened 2015-05-20
Added hardware compatibility list

lzap:


domcleal:
This belongs in iPXE, not in bootdisk as it's specific to iPXE versions.

lzap:
What you mean? I have a user asking - my system does not boot via bootdisk. Is it compatible? But I just learned on the IRC that Full Host is isolinux based, I can add a note how to workaround incompatible hardware.

domcleal:
Whether it's compatible depends on the version of iPXE used.  Wouldn't it be better to keep that information in the iPXE project and reference it?

lzap:
Hmm it's actually here: http://ipxe.org/appnote/hardware_drivers

PR #12, opened 2015-05-27
upd: templates defined for the host are now available on the bootcd

codecap:
### Suggestion

This makes templates defined for the host (operating system) available on the boot cd and
makes possible (for example) to use kickstart unattended installation via the boot cd

Please consider it like a suggestion. I suppose there are more elegant ways to implement the same.

So what do you think?

### DIFF
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -17,9 +17,15 @@ class ForemanBootdisk::ISOGenerator
     # aim to convert filenames to something usable under ISO 9660, update the template to match
     # and then still ensure that the fetch() process stores them under the same name
     files = host.operatingsystem.pxe_files(host.medium, host.architecture, host)
+    # add templates
+    host.operatingsystem.config_templates.each do |t|
+      files << {'unattended/template_'.to_sym => "http://#{host.puppetmaster}/unattended/#{t.template_kind}?token=#{host.token}"}
+    end
     files.map! do |bootfile_info|
       bootfile_info.map do |f|
         suffix = f[1].split('/').last
+        # remove query string from the file name
+        suffix = suffix.split(/[?]/).first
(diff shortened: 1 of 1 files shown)


domcleal:
Seems OK, though I think it would need to use the regular template renderer rather than actually requesting the templates over the API.  The latter could cause a deadlock in a single-process server and also trigger provisioning steps when it shouldn't.

Zugschlus:
Hi domcleal, can you elaborate a bit on this. Unfortunately, I don't have enough clue about foreman's internals to understand what you mean. I am very interested in having this issue resolved. Vlad's patch works fine, but it's  pain to remember applying it after a foreman update, so I am very interested in having the issue resolved in foreman proper.

Zugschlus:
This patch does not work any more in Foreman 1.15:
    2017-09-11 09:25:23 [app] [W] Action failed
     | NoMethodError: undefined method `config_templates' for #<Redhat:0x007fb9c1bfd510>
     | /opt/rh/sclo-ror42/root/usr/share/gems/gems/activemodel-4.2.5.1/lib/active_model/attribute_methods.rb:433:in `method_missing'
     | /opt/theforeman/tfm/root/usr/share/gems/gems/foreman_bootdisk-9.0.0/app/services/foreman_bootdisk/iso_generator.rb:21:in `generate_full_host'
with line 21 being "host.operatingsystem.config_templates.each"

domcleal:
> Hi domcleal, can you elaborate a bit on this.

The code in this PR makes an HTTP request back to the Foreman instance to get the rendered template, but it should use the Foreman template renderer code directly - it isn't normal for a webapp to make HTTP calls to itself when it isn't needed.

(Perhaps call `Host::Managed#render_template`?)

lzap:
What Dominic said, please use the renderer directly similarly how bootdisk does it. Feel free to reopen if you want to push this forward. Cheers.

PR #13, opened 2015-06-10
Fixes #10733: Foreman core changed config to provisioning templates.

ehelms:


db/migrate/20140522185700_change_templatekind_to_bootdisk.rb
```diff
@@ -1,4 +1,6 @@
 class ChangeTemplatekindToBootdisk < ActiveRecord::Migration
+  class ConfigTemplate < ActiveRecord::Base; end
```
domcleal: This is failing to run tests, I suspect because there might be two possible code paths:

1. empty DB, add foreman_bootdisk, then db:migrate.  The config_templates table will be created, this migration runs (note the timestamp/date) using config_templates, then config_templates is renamed to templates.
2. This is what the Jenkins job is doing: empty DB, core only, then db:migrate - config_templates is created and renamed to templates.  Add foreman_bootdisk, this migration runs and expects config_templates.

domcleal:
Also, ticket number is 10773, not 733.

domcleal:
Going to close this in favour of @ares' PR, #14.  Thanks @ehelms.

PR #14, opened 2015-06-11
Fixes #10773 - rename ConfigTemplate to ProvisioningTemplate

ares:


domcleal:
Replaces #13?

ares:
ouch yes, I failed to spot it since the redmine issue was still in new state (wrong ticket number in #13), anyway this renames it hopefully on all places

db/migrate/20140522185700_change_templatekind_to_bootdisk.rb
```diff
@@ -1,12 +1,16 @@
 class ChangeTemplatekindToBootdisk < ActiveRecord::Migration
+  class FakeConfigTemplate < ActiveRecord::Base
+    self.table_name = 'config_templates'
```
domcleal: I'd bet this fails tests for the same reason mentioned in https://github.com/theforeman/foreman_bootdisk/pull/13/files#r32195063.

It's a bit tricky since the ordering of DB migrations from plugins is going to depend on whether they're run together or separately.  Both are valid.

ares:
how about this? (not sure why it was aborted)

domcleal:
Looks good.  Abort was one of the integration tests running on an especially slow slave.

Could you add the min version to engine.rb and README please?

ares:
So I added new version to README, I suppose someone will have to do the release.

domcleal:
Yep, I'll do that soon.  Thanks @ares.

PR #15, opened 2015-08-03
Generic bootdisk without DHCP

glimpsovstar:
Hi,

I have created generic_bootdisk_no_dhcp, for any environment where a user would like to use a generic boot disk without a dhcp.

It requests users to enter IP details.

Can you please test it and let me know what you think?

Regards,
David Joo


app/views/foreman_bootdisk/generic_boot_no_dhcp.erb
```diff
@@ -0,0 +1,41 @@
...
+# was registered with, then prompting for IP settings to configure
+# the interface with.
+#
+
+# loop over net*
+<% (0..32).each do |i| -%>
+:net<%= i %>
+isset ${net<%= i -%>/mac} || goto no_nic
+echo net<%= i -%> is a ${net<%= i -%>/chip} with MAC ${net<%= i -%>/mac}
+echo -n Enter this interface name to boot from it (net<%= i -%>|n):  && read interface
```
domcleal: I'd perhaps put this outside of the loop, so print all interfaces and then ask the user to type in the one they want.

domcleal:
Sorry, I missed this PR!  It works well, I like it - just one inline comment there about the template itself.

If you could also add this to the README and to db/seeds.d/50-bootdisk_templates.rb then it should get created on installation.  I don't know if this should be a different template that's selected from the menu or that should be configured by changing the template name under Settings though.  If the latter then it just needs adding to the DB and the user can reconfigure it.

mmoll:
@glimpsovstar what's the status here?

glimpsovstar:
@domcleal @mmoll Sorry.. somehow the message was gone to wrong folder.
So I am not following what needs to be done. 

The way that I created it, was to add a separate template, so for the users to have 3 different kinds of templates to choose from, depends on their requirement.

domcleal:
Thanks for coming back to it @glimpsovstar.  A few small things need doing:

1) The input/read from the user inside the template could be moved outside of the loop so they see a list of all interfaces and type in the name, instead of going through them one by one.  This should make it easier to use with multiple interfaces.

2) Please edit db/seeds.d/50-bootdisk_templates.rb and add a section that creates this template from the file (when the package/user/installer runs `foreman-rake db:seed`).

3) Please mention the alternative template in the generic section of the README.

lzap:
@glimpsovstar any updates? It is missing just few steps. If you don't plan working on this, let us know and I can perhaps take from here.

app/views/foreman_bootdisk/generic_boot_no_dhcp.erb
glimpsovstar: Hi Dominic,

The reason why this was in there, is to verify that this is the right nic to boot from, by comparing the MAC address.
and in this, the right option will show in "()" as an example "Enter this interface name to boot from it (net0|n): "

Would you still prefer to have it outside of the loop?
If that is the case, I think I would need to put some words around it to remember MAC address and netX from outputs.

lzap: Let me elaborate what (I think) Dominic think. There is a NIC naming convention in iPXE: net0, net1, net2 ... netN. You want the user to enter "netN" to select the proper interface. Now, users will likely identify NIC from MAC addresses. So first use the loop to print out network devices, something like:

```
Found AA:BB:CC:DD:EE:F0 as nic0 
Found AA:BB:CC:DD:EE:F1 as nic1 
Found AA:BB:CC:DD:EE:F2 as nic2
```

And then carry on as usual:

```
Enter interface name to configure: net1
Enter IP address: _
```

app/views/foreman_bootdisk/generic_boot_no_dhcp.erb
```diff
@@ -12,12 +12,16 @@
...
 :net<%= i %>
 isset ${net<%= i -%>/mac} || goto no_nic
-echo net<%= i -%> is a ${net<%= i -%>/chip} with MAC ${net<%= i -%>/mac}
-echo -n Enter this interface name to boot from it (net<%= i -%>|n):  && read interface
-iseq ${interface} net<%= i -%> && goto get_static_ip
+echo Found ${net<%= i -%>/mac} as net<%= i -%> on a ${net<%= i -%>/chip}
 goto net<%= i+1 %>
 <% end -%>
 
+echo -n Enter the interface name to boot from based on above MAC addresses (e.g net0|n):  && read interface
```
lzap: The text is quite long, remember it's only 79 columns. Can you make it shorter?

Also net0|n can be misleading, just "e.g. net0" is fine I guess.

glimpsovstar:
@lzap Done, not it is 75 chars long :)

lzap:
Thanks. Still don't see README and seed changes. Can you add those changes into this PR?

lzap:
Looks like you struggle with git, we want all the three files in this PR which means you want to push them into glimpsovstar repo branch named `patch-1`. There are couple of ways you can achieve this via git but I *think* you are using github editor right?

Do you want me to take from here and create a separate PR?

I created an issue for this feature: http://projects.theforeman.org/issues/23738

glimpsovstar:
Hi Lukas,

If you could take it from there it would be greatly appreciated.

Regards,
David Joo

Powered by djoo’s thumbs.

PR #21, opened 2015-12-08
Add Catalan language

rbuj:


domcleal:
Thanks @rbuj!  Merged as 72b910b.

PR #22, opened 2016-01-19
Fixes #13270 - wait 5 seconds after initialization

lzap:
This slows down generic bootdisk process, but it can avoid problems. I was
thinking introducing a global setting, but since this is a template that can be
cloned and edited, I think this is fine.


domcleal:
You've closed the associated Redmine ticket - are you proposing this for merge or not?

lzap:
Nope, I don't believe that makes any difference, so I am closing this, but keeping the ticket for future reference - we might introduce some kind of global setting with a delay.

PR #24, opened 2016-02-01
Fixes #14794 - added extra information during iPXE boot

lzap:
@domcleal I was wondering if this was a typo (`netX`), according to all docs I
see this should match the device name: http://ipxe.org/cmd/set

This is for discussion first, I haven't tested it yet or created a ticket.

### DIFF
--- a/app/views/foreman_bootdisk/generic_host.erb
+++ b/app/views/foreman_bootdisk/generic_host.erb
@@ -9,6 +9,8 @@
 #
 # Copy this template to customize it, the original is read-only.
 
+echo Foreman Bootdisk: Generic image
+
 # loop over net* until we can get a template
 <% url = bootdisk_chain_url %>
 <% (0..32).each do |i| -%>
--- a/app/views/foreman_bootdisk/host.erb
+++ b/app/views/foreman_bootdisk/host.erb
@@ -14,26 +14,29 @@ bootdisk_raise(N_('Subnet (%s) has no gateway defined'), interface.subnet) if in
 bootdisk_raise(N_('Subnet (%s) has no primary DNS server defined'), interface.subnet) if interface.subnet.dns_primary.nil? || interface.subnet.dns_primary.empty?
 %>
 
+echo Foreman Bootdisk: Host image (<%= @host.name %>)
+
 # loop over net* until the host's MAC matches
 <% (0..32).each do |i| -%>
 :net<%= i %>
 isset ${net<%= i -%>/mac} || goto no_nic
 echo net<%= i -%> is a ${net<%= i -%>/chip} with MAC ${net<%= i -%>/mac}
 iseq ${net<%= i -%>/mac} <%= interface.mac -%> || goto net<%= i+1 %>
-ifopen net<%= i %>
 set idx:int32 <%= i %>
 goto loop_success
 <% end -%>
 
 :loop_success
-echo Configuring net${idx} for static IP address
+echo Configuring net${idx} for static IP address <%= interface.ip %>
 ifopen net${idx}
 set netX/ip <%= interface.ip %>
 set netX/netmask <%= interface.subnet.mask %>
 set netX/gateway <%= interface.subnet.gateway %>
+ifstat net${idx}
 route
 
 # Note, iPXE can only use one DNS server
+echo Using DNS <%= interface.subnet.dns_primary %>
 set dns <%= interface.subnet.dns_primary %>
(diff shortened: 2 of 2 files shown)


app/views/foreman_bootdisk/generic_host.erb
```diff
@@ -9,6 +9,8 @@
 #
 # Copy this template to customize it, the original is read-only.
 
+echo Foreman Bootdisk: Generic image
```
lzap: Adding some extra info because I always get screenshot and I don't know which bootdisk causes the error.

app/views/foreman_bootdisk/generic_host.erb
```diff
@@ -22,5 +24,5 @@ exit 0
 
 :no_nic
 echo Failed to chainload from any network interface
-sleep 30
+sleep 90
```
lzap: Increased this a bit so users can take screenshots or photographs comfortably.

app/views/foreman_bootdisk/host.erb
```diff
@@ -14,26 +14,29 @@ bootdisk_raise(N_('Subnet (%s) has no gateway defined'), interface.subnet) if in
...
 
+echo Foreman Bootdisk: Host image (<%= @host.name %>)
+
 # loop over net* until the host's MAC matches
 <% (0..32).each do |i| -%>
 :net<%= i %>
 isset ${net<%= i -%>/mac} || goto no_nic
 echo net<%= i -%> is a ${net<%= i -%>/chip} with MAC ${net<%= i -%>/mac}
 iseq ${net<%= i -%>/mac} <%= interface.mac -%> || goto net<%= i+1 %>
-ifopen net<%= i %>
```
lzap: I think this `ifopen` is useless, the device is opened below and I believe double opening without properly closing it via `ifclose` might cause issues with some drivers.

app/views/foreman_bootdisk/host.erb
```diff
@@ -14,26 +14,29 @@ bootdisk_raise(N_('Subnet (%s) has no gateway defined'), interface.subnet) if in
...
-echo Configuring net${idx} for static IP address
+echo Configuring net${idx} for static IP address <%= interface.ip %>
 ifopen net${idx}
-set netX/ip <%= interface.ip %>
-set netX/netmask <%= interface.subnet.mask %>
-set netX/gateway <%= interface.subnet.gateway %>
+set net${idx}/ip <%= interface.ip %>
+set net${idx}/netmask <%= interface.subnet.mask %>
+set net${idx}/gateway <%= interface.subnet.gateway %>
+ifstat net${idx}
```
lzap: Here is the actual typo `netX`. This patch fixes it and also adds `ifstat` - additional info about the configured device.

lzap:
According to iPXE sources it looks like `netX` is a valid syntax. But what you think about the double `ifopen`?

app/views/foreman_bootdisk/host.erb
timogoebel: I agree and tested this. Worked for me. :+1:

lzap: Hello, thanks for the feedback. What exactly have you tested? This patch?

app/views/foreman_bootdisk/host.erb
lzap: I searched in the iPXE codebase and I do see "netX" string there with special handling, so perhaps this is correct. I will remove this bit from the patch then. I just can't find any information and the code is not very clear.

timogoebel:
@lzap: What I really love about this PR is the added debug information. That really helps a lot and works fine. :+1: 
I did try removing the double ifopen. You don't need that twice. What I had to leave in though was the "set netX/ip". It did not work with your version.

lzap:
> @lzap: What I really love about this PR is the added debug information. That really helps a lot and works fine. :+1: 
> I did try removing the double ifopen. You don't need that twice. What I had to leave in though was the "set netX/ip". It did not work with your version.

Great, thanks, I will keep netX there.

-- 
Later,
 Lukas #lzap Zapletal

lzap:
Ok I agree, I wanted to kill this patch because it did not work, but I think extra banner is very useful. I have reports from users telling me they are booting generic image while they boot host image and vice versa. @domcleal created ticket, simplified the patch.

lzap:
Also rebased on top of master.

domcleal:
Merged, thanks @lzap.

PR #25, opened 2016-04-18
Fixes #14682 - associate orgs and locs to seeded templates

ares:


domcleal:
Merged as ab0c39d, thanks @ares.

PR #26, opened 2016-04-25
fixes #14796 - show bootdisk fields after new host page reload

timogoebel:


domcleal:
Merged as 22f6e22, thanks @timogoebel.

PR #27, opened 2016-05-06
Fixes #23713 - static credentials fail over support

lzap:
This is a patch based on an idea by Morgan Weetman and David Joo to enter
network credentials manually when DHCP is not available. I actually added this
as a fallback mechanism when DHCP attempt fails or when Foreman returns 404 on
MAC address.

In that case, a set of queries is showed and then another boot attempt is done.
If that fails, user is dropped to shell (I also changed that to be the default
behavior for both image types which seems better than useless booting loop).

I tested this in both scenarios (no DHCP response, Foreman gives 404) and it
works great. In the latter case, IP address is already pre-filled so user can
change it. We should consider adding foreman URL there as a variable so user
could verify it (and modify) too just in case.

### DIFF
--- a/app/views/foreman_bootdisk/generic_host.erb
+++ b/app/views/foreman_bootdisk/generic_host.erb
@@ -21,8 +21,26 @@ dhcp net<%= i %> || goto net<%= i+1 %>
 chain <%= url %>${net<%= i -%>/mac} || goto net<%= i+1 %>
 exit 0
 <% end -%>
+echo Failed to chainload from any network interface, fallback to static.
+ifstat
 
 :no_nic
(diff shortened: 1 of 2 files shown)


lzap:
ipxe-static

lzap:
This patch is similar to https://github.com/theforeman/foreman_bootdisk/pull/15 but it's a failover case. Could be useful, this is now ready for review @timogoebel

lzap:
Test failures seems unrelated, v12 is out, this is only failed scenario case - no risk.

timogoebel:
Thanks, @lzap.

> no risk

I'll just make you fix it if it breaks something :-P

PR #28, opened 2016-05-06
Fixes #14948 - Display select boxes on settings where we can

sean797:


app/helpers/concerns/foreman_bootdisk/settings_helper_extensions.rb
```diff
@@ -0,0 +1,29 @@
...
+      alias_method_chain :value, :bootdisk
+    end
+
+    def value_with_bootdisk(setting)
+      return value_without_bootdisk(setting) unless [
+        'bootdisk_generic_host_template',
+        'bootdisk_host_template'].include?(setting.name)
+
+      case setting.name
+      when "bootdisk_generic_host_template"
```
domcleal: It appears that https://github.com/theforeman/foreman/commit/ad603e4023c5400c51674f1fa3635565f7b3b552#diff-8c3f2a547e7f7d66e1756971a1ef20c2R275 adds some support for passing a collection into the creation of Settings, so there's no need to extend Foreman's helper here. This could be moved to the Setting::Bootdisk class.

sean797: That makes it much cleaner, thanks @domcleal :+1:

app/models/setting/bootdisk.rb
```diff
@@ -10,8 +10,8 @@ def self.load_defaults
       [
         self.set('bootdisk_ipxe_dir', N_('Path to directory containing iPXE images'), ipxe),
         self.set('bootdisk_syslinux_dir', N_('Path to directory containing syslinux images'), syslinux),
-        self.set('bootdisk_host_template', N_('iPXE template to use for host-specific boot disks'), 'Boot disk iPXE - host'),
-        self.set('bootdisk_generic_host_template', N_('iPXE template to use for generic host boot disks'), 'Boot disk iPXE - generic host'),
+        self.set('bootdisk_host_template', N_("iPXE template to use for host-specific boot disks"), '', 'bootdisk_host_template', nil, { :collection => Hash[ProvisioningTemplate.where(:template_kind => TemplateKind.where(:name => 'iPXE')).map{|temp| [temp[:name], temp[:name]]}]}),
```
domcleal: The template kind should be Bootdisk.

app/models/setting/bootdisk.rb
```diff
@@ -10,8 +10,8 @@ def self.load_defaults
       [
         self.set('bootdisk_ipxe_dir', N_('Path to directory containing iPXE images'), ipxe),
         self.set('bootdisk_syslinux_dir', N_('Path to directory containing syslinux images'), syslinux),
-        self.set('bootdisk_host_template', N_('iPXE template to use for host-specific boot disks'), 'Boot disk iPXE - host'),
-        self.set('bootdisk_generic_host_template', N_('iPXE template to use for generic host boot disks'), 'Boot disk iPXE - generic host'),
+        self.set('bootdisk_host_template', N_("iPXE template to use for host-specific boot disks"), '', 'bootdisk_host_template', nil, { :collection => Hash[ProvisioningTemplate.where(:template_kind => TemplateKind.where(:name => 'Bootdisk')).map{|temp| [temp[:name], temp[:name]]}]}),
```
domcleal: The third parameter (the default value) should stay I think, else the plugin won't be correctly configured without the templates being explicitly set.

domcleal: and keep the single quotes, no need to change them.

domcleal:
If you update the PR, please leave a comment so I know.

sean797:
How's that?

domcleal:
Merged as 1211696, thanks @sean797.

PR #29, opened 2016-05-25
fixes #15173 - vmware new cdrom was moved to core

timogoebel:
This was moved to core in https://github.com/theforeman/foreman/pull/3540


domcleal:
Thanks!

PR #31, opened 2016-07-22
add host templates to iso before building

Zugschlus:
Hi,

pardon me for directly jumping in with a one-liner pull requests, but I didn't find issues in the theforeman/foreman_bootdisk so I'm going to file a pull request.

This pull request will cause the host templates to be included in the .iso, which is necessary at my work site to successfully build a host. I don't have any clue whether this is helpful to the general public though. Please feel free to reject my suggestion if you don't find it helpful.

Greetings
Marc


domcleal:
#12 does an almost identical thing, but the same comment there applies to this - it should be rendering the templates internally, not making HTTP calls back to itself.

Zugschlus:
yes, this is identical. closing.

PR #32, opened 2016-11-09
Fixed Full Host image generation on 1.13

lzap:
PXELoader flag in 1.13 broke Full Host generation as the parameter is not
passed. We can fix this here, or I can provide default parameter in core to
remain compatibility (but it can bite). I prefer being explicit.

Anyway, this is bad regression, any chance to squeeze this into 1.13 plugin? Do
you want me to do a ticket?

### DIFF
--- a/test/unit/iso_generator_test.rb
+++ b/test/unit/iso_generator_test.rb
@@ -1,0 +1,21 @@
+require 'test_plugin_helper'
+
+class ForemanBootdisk::IsoGeneratorTest < ActiveSupport::TestCase
+  include ForemanBootdiskTestHelper
+  setup :setup_bootdisk
+  setup :setup_org_loc
+  setup :setup_subnet
+  setup :setup_host
+
+  setup do
+    @host.build = true
+  end
+
+  test "full host image generation generates via PXELinux type" do
+    @host.operatingsystem.expects(:pxe_files).returns([])
+    @host.operatingsystem.expects(:kernel).returns("kernel")
+    @host.operatingsystem.expects(:initrd).returns("initrd")
+    ForemanBootdisk::ISOGenerator.expects(:generate).with({:isolinux => nil, :files => []}, anything)
+    ForemanBootdisk::ISOGenerator.generate_full_host(@host)
+  end
+end
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -10,7 +10,7 @@ class ForemanBootdisk::ISOGenerator
   def self.generate_full_host(host, opts = {}, &block)
     raise ::Foreman::Exception.new(N_('Host is not in build mode, so the template cannot be rendered')) unless host.build?
 
-    tmpl = host.send(:generate_pxe_template)
+    tmpl = host.send(:generate_pxe_template, :PXELinux)
     raise ::Foreman::Exception.new(N_('Unable to generate disk template: %s'), host.errors.full_messages.to_sentence) if tmpl == false
 
     # pxe_files and filename conversion is utterly bizarre
--- a/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
+++ b/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
@@ -1,6 +1,7 @@
 require 'test_plugin_helper'
 
 class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
+  include ForemanBootdiskTestHelper
   setup :setup_bootdisk
 
   test "should generate generic image" do
@@ -12,6 +13,7 @@ class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
   end
 
   describe "#host" do
+    setup :setup_referer
     setup :setup_host_env
 
     test "should generate host image" do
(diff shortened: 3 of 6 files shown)


domcleal:
Can you add a unit test of `generate_full_host` please? It's OK to stub/expect `generate` itself.

lzap:
Sure, I had to slightly refactor the helper in order to use it outside controller context.

test/unit/iso_generator_test.rb
```diff
@@ -0,0 +1,20 @@
...
+  setup :setup_org_loc
+  setup :setup_subnet
+  setup :setup_host
+
+  setup do
+    @host.build = true
+  end
+
+  test "full host image generation generates via PXELinux type" do
+    @host.expects(:generate_pxe_template).with(:PXELinux)
```
domcleal: Doesn't stubbing this undermine the purpose of the test, to make sure it actually works? If it's stubbed expecting an argument and the method signature changes, the test will never fail.

lzap: Well my goal was to test only PXELinux option in `generate_pxe_template`. Anyway, if I understand you correctly, you want this test to be complete so changed it accordingly.

Also added two more tests on the filename part, which looks quite fragile.

test/unit/iso_generator_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    @host.operatingsystem.expects(:pxe_files).returns(boot_files)
+    ForemanBootdisk::ISOGenerator.expects(:generate).with({
+      :isolinux => 'Fake kernel line BOOT/NUXATOMIC_7_3_X86_64_VMLINUZ',
+      :files => [[['BOOT/NUXATOMIC_7_3_X86_64_VMLINUZ', '/tmp/pxeboot/vmlinuz']],
+        [['BOOT/ATOMIC_7_3_X86_64_INITRD_IMG', '/tmp/pxeboot/initrd.img']]]}, anything)
+    ForemanBootdisk::ISOGenerator.generate_full_host(@host)
+  end
+
+  test "full host image generation generates via PXELinux type" do
+    @host.expects(:generate_pxe_template).with(:PXELinux).returns("Fake content")
```
domcleal: > What exactly will cause the test to fail? Are you commenting on line 15 or 17?

This was on line 15, before you changed it. I'm saying that a change in the `generate_pxe_template` signature will _not_ cause the test to fail. This shouldn't be stubbed, it should be calling the real method.

test/unit/iso_generator_test.rb
```diff
@@ -0,0 +1,48 @@
...
+  setup :setup_bootdisk
+  setup :setup_org_loc
+  setup :setup_subnet
+  setup :setup_host
+
+  setup do
+    @host.build = true
+  end
+
+  test "generate_full_host creates with ISO-compatible file names" do
```
domcleal: These look unrelated to the fix, please move them into their own PR.

lzap:
Rebased. Removed the two guys:

```ruby
  test "generate_full_host creates with ISO-compatible file names" do
    @host.expects(:generate_pxe_template).with(:PXELinux).returns("Fake kernel line boot/Debian-8.1-x86_64-vmlinuz")
    boot_files = [
      {:"boot/Debian-8.1-x86_64"=>"/tmp/pxeboot/vmlinuz"},
      {:"boot/Debian-8.1-x86_64"=>"/tmp/pxeboot/initrd.img"}
    ]
    @host.operatingsystem.expects(:pxe_files).returns(boot_files)
    ForemanBootdisk::ISOGenerator.expects(:generate).with({
      :isolinux => 'Fake kernel line BOOT/DEBIAN_8_1_X86_64_VMLINUZ',
      :files => [[['BOOT/DEBIAN_8_1_X86_64_VMLINUZ', '/tmp/pxeboot/vmlinuz']],
      [['BOOT/DEBIAN_8_1_X86_64_INITRD_IMG', '/tmp/pxeboot/initrd.img']]] }, anything)
    ForemanBootdisk::ISOGenerator.generate_full_host(@host)
  end

  test "generate_full_host creates with ISO-compatible long file names" do
    @host.expects(:generate_pxe_template).with(:PXELinux).returns("Fake kernel line boot/RedHatEnterpriseLinuxAtomic-7.3-      +x86_64-vmlinuz")
    boot_files = [
      {:"boot/RedHatEnterpriseLinuxAtomic-7.3-x86_64"=>"/tmp/pxeboot/vmlinuz"},
      {:"boot/RedHatEnterpriseLinuxAtomic-7.3-x86_64"=>"/tmp/pxeboot/initrd.img"}
    ]
    @host.operatingsystem.expects(:pxe_files).returns(boot_files)
    ForemanBootdisk::ISOGenerator.expects(:generate).with({
      :isolinux => 'Fake kernel line BOOT/NUXATOMIC_7_3_X86_64_VMLINUZ',
      :files => [[['BOOT/NUXATOMIC_7_3_X86_64_VMLINUZ', '/tmp/pxeboot/vmlinuz']],
        [['BOOT/ATOMIC_7_3_X86_64_INITRD_IMG', '/tmp/pxeboot/initrd.img']]]}, anything)
    ForemanBootdisk::ISOGenerator.generate_full_host(@host)
  end
```

I will file PR once this is merged, filed: http://projects.theforeman.org/issues/18823

lzap:
Thanks here are the missing tests: https://github.com/theforeman/foreman_bootdisk/pull/38

PR #33, opened 2016-11-09
Fixes #4413 - full host image EFI support

lzap:
By default this patch should not change anything, but if user copies
`efiboot.img` and `grubx64.efi` into a directory and sets this in the settings,
Full Host image will be created with EFI in mind (as hybrid ISO as well). So
the ISO will work both on BIOS and EFI systems.

For RHEL systems, the file can be only find in the kickstart tree:

http://ftp.fi.muni.cz/pub/linux/centos/7/os/x86_64/images/

Grub2 file can be found in grub2-efi package.

Not sure about Debian based distros, this will work there as well.

### DIFF
--- a/app/models/setting/bootdisk.rb
+++ b/app/models/setting/bootdisk.rb
@@ -1,4 +1,6 @@
 class Setting::Bootdisk< ::Setting
+  BLANK_ATTRS << "bootdisk_efiboot_dir"
+
   def self.load_defaults
     return unless ActiveRecord::Base.connection.table_exists?('settings')
     return unless super
@@ -13,6 +15,7 @@ class Setting::Bootdisk< ::Setting
         self.set('bootdisk_ipxe_dir', N_('Path to directory containing iPXE images'), ipxe, N_('iPXE directory')),
         self.set('bootdisk_isolinux_dir', N_('Path to directory containing isolinux images'), isolinux, N_('ISOLINUX directory')),
         self.set('bootdisk_syslinux_dir', N_('Path to directory containing syslinux images'), syslinux, N_('SYSLINUX directory')),
+        self.set('bootdisk_efiboot_dir', N_('Path to directory containing efiboot.img and grubx64.efi (leave empty for BIOS only)'), '', N_('EFI directory')),
         self.set('bootdisk_host_template', N_('iPXE template to use for host-specific boot disks'), 'Boot disk iPXE - host', N_('Host image template'), nil, :collection => templates),
         self.set('bootdisk_generic_host_template', N_('iPXE template to use for generic host boot disks'), 'Boot disk iPXE - generic host', N_('Generic image template'), nil, :collection => templates),
         self.set('bootdisk_mkiso_command', N_('Command to generate ISO image, use genisoimage or mkisofs'), 'genisoimage', N_('ISO generation command')),
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -10,8 +10,9 @@ class ForemanBootdisk::ISOGenerator
   def self.generate_full_host(host, opts = {}, &block)
     raise ::Foreman::Exception.new(N_('Host is not in build mode, so the template cannot be rendered')) unless host.build?
 
-    tmpl = host.send(:generate_pxe_template)
-    raise ::Foreman::Exception.new(N_('Unable to generate disk template: %s'), host.errors.full_messages.to_sentence) if tmpl == false
+    pxelinux_template = host.send(:generate_pxe_template, :PXELinux)
+    pxegrub2_template = host.send(:generate_pxe_template, :PXEGrub2) if Setting[:bootdisk_efiboot_dir].present?
+    raise ::Foreman::Exception.new(N_('Unable to generate disk template: %s'), host.errors.full_messages.to_sentence) if pxelinux_template == false
 
     # pxe_files and filename conversion is utterly bizarre
     # aim to convert filenames to something usable under ISO 9660, update the template to match
@@ -21,13 +22,14 @@ class ForemanBootdisk::ISOGenerator
       bootfile_info.map do |f|
         suffix = f[1].split('/').last
         iso_f0 = iso9660_filename(f[0].to_s + '_' + suffix)
-        tmpl.gsub!(f[0].to_s + '-' + suffix, iso_f0)
+        pxelinux_template.gsub!(f[0].to_s + '-' + suffix, iso_f0)
+        pxegrub2_template.gsub!(f[0].to_s + '-' + suffix, '/' + iso_f0) if pxegrub2_template
         ForemanBootdisk.logger.debug("Boot file #{iso_f0}, source #{f[1]}")
         [iso_f0, f[1]]
       end
     end
 
-    generate(opts.merge(:isolinux => tmpl, :files => files), &block)
+    generate(opts.merge(:isolinux => pxelinux_template, :grub => pxegrub2_template, :files => files), &block)
   end
 
   def self.generate(opts = {}, &block)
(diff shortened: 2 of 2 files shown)


domcleal:
Where is the support for the main images, not full host images?

lzap:
There is none, do you think it will be possible to chainload ipxe from grub2-efi? I thought EFI environment is locked and you can't exit by loading non-EFI binary.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -10,8 +10,9 @@ class ForemanBootdisk::ISOGenerator
   def self.generate_full_host(host, opts = {}, &block)
     raise ::Foreman::Exception.new(N_('Host is not in build mode, so the template cannot be rendered')) unless host.build?
 
-    tmpl = host.send(:generate_pxe_template)
-    raise ::Foreman::Exception.new(N_('Unable to generate disk template: %s'), host.errors.full_messages.to_sentence) if tmpl == false
+    pxelinux_template = host.send(:generate_pxe_template, :PXELinux)
```
domcleal: This is from #32, needs completing first.

app/models/setting/bootdisk.rb
```diff
@@ -13,6 +15,7 @@ def self.load_defaults
         self.set('bootdisk_ipxe_dir', N_('Path to directory containing iPXE images'), ipxe, N_('iPXE directory')),
         self.set('bootdisk_isolinux_dir', N_('Path to directory containing isolinux images'), isolinux, N_('ISOLINUX directory')),
         self.set('bootdisk_syslinux_dir', N_('Path to directory containing syslinux images'), syslinux, N_('SYSLINUX directory')),
+        self.set('bootdisk_efiboot_dir', N_('Path to directory containing efiboot.img and grubx64.efi (leave empty for BIOS only)'), '', N_('EFI directory')),
```
domcleal: 1. Missing from README.
1. Why default to an empty path? Isn't this known for most OSes?

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -75,13 +77,30 @@ def self.generate(opts = {}, &block)
...
+      mkiso_command = "#{Setting[:bootdisk_mkiso_command]} -o #{iso} -iso-level 2 -b isolinux.bin -c boot.cat -no-emul-boot -boot-load-size 4 -boot-info-table #{efiopts} #{File.join(wd, 'build')}"
+      unless system(mkiso_command)
+        raise ::Foreman::Exception.new(N_("ISO build failed") + ': ' + mkiso_command)
       end
 
       # Make the ISO bootable as a HDD/USB disk too
-      unless system("isohybrid", iso)
-        raise ::Foreman::Exception.new(N_("ISO hybrid conversion failed"))
+      unless system(isohybrid_command)
+        raise ::Foreman::Exception.new(N_("ISO hybrid conversion failed") + ': ' + isohybrid_command)
```
domcleal: Don't append strings to messages, pass them as arguments to Foreman::Exception.

domcleal:
In the ticket you stated that iPXE has support for UEFI booting, so can't you just boot that directly?

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -10,8 +10,9 @@ class ForemanBootdisk::ISOGenerator
   def self.generate_full_host(host, opts = {}, &block)
     raise ::Foreman::Exception.new(N_('Host is not in build mode, so the template cannot be rendered')) unless host.build?
 
-    tmpl = host.send(:generate_pxe_template)
-    raise ::Foreman::Exception.new(N_('Unable to generate disk template: %s'), host.errors.full_messages.to_sentence) if tmpl == false
+    pxelinux_template = host.send(:generate_pxe_template, :PXELinux)
+    pxegrub2_template = host.send(:generate_pxe_template, :PXEGrub2) if Setting[:bootdisk_efiboot_dir].present?
+    raise ::Foreman::Exception.new(N_('Unable to generate disk template: %s'), host.errors.full_messages.to_sentence) if pxelinux_template == false
```
domcleal: Missing a corresponding check for failures to render the PXEGrub2 template

lzap:
> In the ticket you stated that iPXE has support for UEFI booting, so can't you just boot that directly?

I will try that, of course I can include it in the patch. Let me test this first.

> Note that the ticket is also for general UEFI support, I don't wish to support full disk only, this is not a recommended way of using the plugin.

Well, according to my numbers, it is very popular way of using the plugin because of limited iPXE driver capability.

lzap:
There's a problem with chainloading iPXE.EFI from Grub2, loading it directly would mean we need to compile and embed the script into it. Asked on the ipxe-dev list about this: http://lists.ipxe.org/pipermail/ipxe-devel/2016-December/thread.html#5279 (for the record)

lzap:
Ok so it's confirmed, in EFI mode it is not possible to inject the script, the only way of doing this is embedding the template compile time or at least embedding chainloading of a template that way.

```
#!ipxe
prompt --key 0x02 --timeout 2000 Press Ctrl-B for the iPXE command line... && shell ||
chain file:boot.ipxe
```

This was suggested in http://forum.ipxe.org/showthread.php?tid=8244 and we'd need to build the package ourselves, or tell the Foreman user to do this. Both seem weird to me and usability of this is low.

Therefore I asked our bootloaders team to evaluate how much work would it be to implement the chainloading via command line: https://bugzilla.redhat.com/show_bug.cgi?id=1401491

We have two opinions now, either wait or implement full host image only. Any other ideas?

lzap:
Just for the record, there was a patch to accept command line options in iPXE.EFI but that does not help as there is no way of passing that from Grub2. The only option is to build FAT compatible image with iPXE.EFI built with embedded chain script. That means carrying ipxe build in our repositories. Here is an example how to create the FAT image:

https://github.com/openstack/diskimage-builder/blob/master/elements/iso/cleanup.d/100-build-iso

davidvoit:
Hi all,

as we are needing this at work, I looked into it. Do get ipxe working it's needs to fake a recent linux behaviour and needs to implement the "handover protocol". Details can be found in patch 20 "Add support for linuxefi" in the grub2 srpm (I looked at fedora rawhide package). I will try to get my hands dirty with the ipxe code, maybe I can get a proof of concept patch running.

Greetings,
David

davidvoit:
Btw. after getting ipxe with handover to boot (handover function is not working yet - I struggle with the crazy elf to pe function part, I need the address to continue... But no longer magic checksum errors :-) )

Two things struck me:
1.) in your patch copying grub is not necessary, this is already included in the efi img. And maybe modifiy efi img is the cleaner solution. Else you overlay the fat entry using the iso9660 file. I got this entry: _GRUB_CF.UN_ but not tested your script, I did this by hand
2.) why not use the inbuilt http support of grub, this use the efinet protocol, so the driver stuff of ipxe in classic bios mode is not necessary... Yes this means a new template, but no change in ipxe is necessary. I think this is the cleaner solution.

lzap:
@davidvoit I don't insist on Grub2 at all, it was just my first idea to do it this way. If you have an idea how to do EFI for iPXE-based images go ahead. Do you want to try the patch for bootdisk or explain to me how this is supposed to work? Because I don't exactly understand what you mean.

davidvoit:
@lzap 
Hi Lukáš,
I think you can't avoid grub2. I would just use grub2!

Something like this:
```
set timeout=5

menuentry 'Linux diskless' --class os {
     insmod net
     insmod efinet
     insmod tftp
     insmod http
     insmod gzio
     insmod part_gpt
     insmod efi_gop
     insmod efi_uga

     # http server in my network
     set net_default_server=192.168.100.2

     # net_bootp

     # ok let's assign a static address for now
     net_add_addr eno0 efinet0 192.168.100.9

     echo 'Network status: '
     net_ls_cards
     net_ls_addr
     net_ls_routes

     echo 'Loading Linux ...'
     linuxefi (http)/boot/linux

     echo 'Loading initial ramdisk ...'
     initrdefi (http)/boot/initrd
}
```
From: https://github.com/openSUSE/kiwi/wiki/Setup-PXE-boot-with-EFI-Using-GRUB2

If you are using iPXE in uefi mode you can't (currently) load your script file!

If this is ok, do it without iPXE than I can take a look about the code, about this weekend.

Greetings
David

lzap:
@domcleal UEFI support for host/generic images can't be easily done, would you accept the original patch with only Full Host disk support if I'd rebase it? Thanks.

domcleal:
Sorry, I think this is critical - I'm not willing to support full host images, and certainly not recommend that they're used for UEFI hosts. Please update the PR when you can support the primary generic images, or consider putting this code into a different project.

PR #34, opened 2016-12-01
Fixes #17552 - set default api version to v2

swapab:
Apipie doc falsely states url(s) for bootdisk API calls.
Example: `/api/generic` should have been `/bootdisk/api/v2/generic`.
I wasn't sure how to test it. Should the API url be tested from foreman?

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,7 +10,7 @@ ForemanBootdisk::Engine.routes.draw do
   end
 
   namespace :api, :defaults => {:format => 'json'} do
-    scope "(:apiv)", :module => :v2, :defaults => {:apiv => 'v2'}, :apiv => /v1|v2/, :constraints => ApiConstraints.new(:version => 2) do
+    scope "(:apiv)", :module => :v2, :defaults => {:apiv => 'v2'}, :apiv => /v1|v2/, :constraints => ApiConstraints.new(:version => 2, :default => true) do
       get 'generic', :to => 'disks#generic'
       constraints(:id => /[^\/]+/) do
         get 'hosts/:id', :to => 'disks#host'
--- a/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
+++ b/test/functional/foreman_bootdisk/api/v2/disks_controller_test.rb
@@ -12,9 +12,7 @@ class ForemanBootdisk::Api::V2::DisksControllerTest < ActionController::TestCase
   end
 
   describe "#host" do
-    setup :setup_org_loc
-    setup :setup_subnet
-    setup :setup_host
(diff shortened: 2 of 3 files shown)


domcleal:
The URL is correct if the version header is passed - please instead fix the default API version to match Foreman, instead of changing the documentation URL. The question (which I didn't understand) should not be in the commit message.

swapab:
@domcleal Thanks for the review.
Updated the patch to set default API version to _v2_ using ApiConstraints from [Foreman](https://github.com/theforeman/foreman/blob/develop/app/services/api_constraints.rb).

The question I asked was about testing the API URL (integration testing most likely). How is it usually done via foreman or within the plugin itself or not required at all since functional tests are in place ?

Didn't found any tests for api_constraints.rb as well. Should I add them in a different foreman patch ?

domcleal:
You can add integration tests if you wish, or use Rails' routing test helpers. If you're doing that, I'll set this to WoC.

swapab:
Okay I'll test using [Rails routing helpers](http://api.rubyonrails.org/v4.2.7.1/classes/ActionDispatch/Assertions/RoutingAssertions.html).

Alongside while testing entire test suite of foreman_bootdisk I am getting a bunch of errors. Thought it might be familiar to you

`centos7-devel :: ~/foreman ‹develop*› » rake test:foreman_bootdisk`

```
.EEEE..EEE.......................

Finished in 6.314247s, 5.2263 runs/s, 9.5023 assertions/s.

  1) Error:
ForemanBootdisk::Api::V2::DisksControllerTest::#host#test_0001_should generate host image:
ActiveRecord::RecordInvalid: Validation failed: Download policy must be one of the following: immediate, on_demand, background, inherit
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:11:in `block in transaction'
    test_after_commit (1.1.0) lib/test_after_commit/database_statements.rb:5:in `transaction'
    /home/vagrant/katello/app/lib/katello/lazy_accessor.rb:87:in `save!'
    /home/vagrant/foreman_bootdisk/test/test_plugin_helper.rb:29:in `setup_subnet'
```

Full trace : https://gist.github.com/swapab/abe825973a6db64d7408ab0fb30050f8

---

Any idea how to fix those ?

domcleal:
No, that does not look like foreman_bootdisk, disable your other plugins.

swapab:
@domcleal Test cases added : https://github.com/theforeman/foreman_bootdisk/pull/34/commits/403315b7e098761734047c0fe1174c569ca71907#diff-3be020c4f93f48ef73b2df6232dfddd1

domcleal:
Merged, thanks @swapab.

PR #35, opened 2016-12-02
Fixes #17496 - disable Boot disk button for non-intel arch

swapab:
disabled-boot-disk

### DIFF
--- a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
+++ b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
@@ -6,20 +6,34 @@ module ForemanBootdisk::HostsHelperExt
   end
 
   def host_title_actions_with_bootdisk(*args)
-    title_actions(
-      button_group(
-        select_action_button(_('Boot disk'), {:class => 'btn btn-group'},
-          display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'la'),
-          display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'la'),
-          content_tag(:li, "", :class => "divider"),
-          display_bootdisk_link_if_authorized(_("Generic image"), {:controller => 'foreman_bootdisk/disks', :action => 'generic'}, :class=>'la'),
-          display_bootdisk_for_subnet,
-          content_tag(:li, "", :class => "divider"),
-          display_bootdisk_link_if_authorized(_("Help"), {:controller => 'foreman_bootdisk/disks', :action => 'help'}, :class=>'la')
+    if @host.bootdisk_downloadable?
+      title_actions(
+        button_group(
+          select_action_button(_('Boot disk'), {:class => 'btn btn-group'},
+            display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'la'),
+            display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'la'),
+            content_tag(:li, "", :class => "divider"),
+            display_bootdisk_link_if_authorized(_("Generic image"), {:controller => 'foreman_bootdisk/disks', :action => 'generic'}, :class=>'la'),
+            display_bootdisk_for_subnet,
+            content_tag(:li, "", :class => "divider"),
+            display_bootdisk_link_if_authorized(_("Help"), {:controller => 'foreman_bootdisk/disks', :action => 'help'}, :class=>'la')
+          )
         )
       )
-    )
+    else
+      bootdisk_button_disabled
+    end
+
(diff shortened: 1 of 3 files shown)


app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -6,20 +6,33 @@ module ForemanBootdisk::HostsHelperExt
...
-        select_action_button(_('Boot disk'), {:class => 'btn btn-group'},
-          display_bootdisk_link_if_authorized(_("Host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'host', :id => @host}, :class=>'la'),
-          display_bootdisk_link_if_authorized(_("Full host '%s' image") % @host.name.split('.')[0], {:controller => 'foreman_bootdisk/disks', :action => 'full_host', :id => @host}, :class=>'la'),
-          content_tag(:li, "", :class => "divider"),
-          display_bootdisk_link_if_authorized(_("Generic image"), {:controller => 'foreman_bootdisk/disks', :action => 'generic'}, :class=>'la'),
-          display_bootdisk_for_subnet,
-          content_tag(:li, "", :class => "divider"),
-          display_bootdisk_link_if_authorized(_("Help"), {:controller => 'foreman_bootdisk/disks', :action => 'help'}, :class=>'la')
-        )
+        link_to(_("Boot disk"), '#', :disabled => true, :class => 'btn btn-default', :title => "Bootdisk download not available for #{@host.architecture.name} architecture")
```
domcleal: 1. The test failures likely come from this #name call.
2. Missing string extraction.
3. "Boot disk" is two words.

swapab: @domcleal Patch updated
Above 1/2/3 done.

app/models/concerns/foreman_bootdisk/host_ext.rb
```diff
@@ -33,11 +33,17 @@ def bootdisk?
     managed? && bootdisk_build? && SETTINGS[:unattended]
   end
 
+  def bootdisk_downloadable?
+    ['ia32', 'x64'].include?(architecture.try(:bootfilename_efi))
```
domcleal: Why use `bootfilename_efi` rather than the arch name?

app/models/concerns/foreman_bootdisk/host_ext.rb
```diff
@@ -33,11 +33,17 @@ def bootdisk?
...
+
   def validate_media_with_bootdisk?
     validate_media_without_bootdisk? || (managed && bootdisk_build? && build?)
   end
 
   def can_be_built_with_bootdisk?
     can_be_built_without_bootdisk? || (managed? and SETTINGS[:unattended] and bootdisk_build? and !build?)
   end
+
+
```
domcleal: Unnecessary.

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -6,20 +6,34 @@ module ForemanBootdisk::HostsHelperExt
   end
 
   def host_title_actions_with_bootdisk(*args)
+    if @host.architecture.blank? || @host.bootdisk_downloadable?
```
domcleal: I think accepting `architecture.blank?` is OK, but shouldn't this be in `bootdisk_downloadable?`, otherwise it will indicate false when it _is_ permitted.

test/unit/concerns/host_test.rb
```diff
@@ -40,4 +40,27 @@ class ForemanBootdisk::HostTest < ActiveSupport::TestCase
       assert_equal false, h.bootdisk?
     end
   end
+
+  context "#bootdisk_downloadable?" do
```
domcleal: Needs test for blank value.

app/models/concerns/foreman_bootdisk/host_ext.rb
swapab: @domcleal [`Architecture#bootfilename_efi`](https://github.com/theforeman/foreman/blob/develop/app/models/architecture.rb#L19-L30) gives a more refined result of the arch. Although I could have checked for '86' in `host.arch` but to avoid generic set 'ia32' and 'x64' looks to be more prominent architectures belonging to intel.

swapab: Okay. It seemed closely related to `architecture#name` so I thought to use it rather than writing a new predicate method.
I suppose `architecture#intel?` would be meaningful.

```ruby
def intel?
  /i.86|x86[_-]64/ =~ name
end
```

swapab: I am actually not well versed with OS and it's underlying architecture, just know basics. So I am assuming intel doesn't have anything more than those architecture names(i386 for 32 bit and x86_64 is 64 bit)

app/models/concerns/foreman_bootdisk/host_ext.rb
```diff
@@ -33,6 +33,14 @@ def bootdisk?
     managed? && bootdisk_build? && SETTINGS[:unattended]
   end
 
+  def bootdisk_downloadable?
+    architecture.blank? || intel_arch?
+  end
+
+  def intel_arch?
```
domcleal: No need to extend the model in this plugin, it's unnecessary complexity when it's only used by the UI helper.

domcleal:
It's unclear to me if you're moving code around further, please indicate when it's actually ready for re-review.

swapab:
@domcleal 
All changes have been done from my side. Good to review.

domcleal:
Merged, thanks @swapab.

PR #36, opened 2017-01-20
Fixes #17893 - Do not preview bootdisk templates

xprazak2:
needs [this](https://github.com/theforeman/foreman/pull/4211) to work


domcleal:
Why should the preview be disabled?

xprazak2:
The preview ends up in error, I based this on @lzap's [comment](https://bugzilla.redhat.com/show_bug.cgi?id=1393848#c4) saying that preview is not relevant here.

domcleal:
The error should be fixed then, this PR makes little sense to me.

lzap:
I think we should prevent association than preview, if we want to fix it like that. I think this could be simply better documentation or similar.

PR #41, opened 2017-05-25
Full Host disk DHCP reservation docs

lzap:
This is no longer the case with 1.15 templates:

https://github.com/theforeman/community-templates/pull/326/files

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ TFTP settings are needed.
   <tr>
     <td>Full host image</td>
     <td>No</td>
-    <td>Yes</td>
+    <td>No (*)</td>
     <td>No</td>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -144,6 +144,8 @@ TFTP settings are needed.
   </tr>
 </table>
 
+(*) - Foreman 1.14 or older do require DHCP reservation
```
lzap: Yeah bad wording, fixed. To confirm: Nowdays neither DHCP service nor DHCP reservation is required, PXELinux template will create Anaconda configuration with static IP just fine.

domcleal:
Merged, thanks.

PR #44, opened 2017-10-09
Fixes #21106 - gateway-less support for host images

lzap:
Gateway is not always needed.


app/views/foreman_bootdisk/host.erb
```diff
@@ -31,7 +30,9 @@ echo Configuring net${idx} for static IP address <%= interface.ip %>
 ifopen net${idx}
 set netX/ip <%= interface.ip %>
 set netX/netmask <%= interface.subnet.mask %>
+<% if interface.subnet.gateway %>
```
timogoebel: `.present?` to also catch an empty value?

timogoebel:
@lzap: Moved the Redmine ticket to Bootdisk. When you fix my other comment, this is ready.

app/views/foreman_bootdisk/host.erb
lzap: Done!

timogoebel:
@domcleal: Anything to add?

ares:
@timogoebel any chance we could get this in? it seems it was waiting only on pressing the button :-)

timogoebel:
Merged, thanks @lzap and @ares.

ares:
Awesome, thanks!

PR #45, opened 2017-10-09
fixes #21257 - tests are rails 5 compatible

timogoebel:
The test failure was introduced by:
https://github.com/theforeman/foreman/commit/a19834d8d32bd12024f757f9ec3101a918c1481e


domcleal:
Thanks @timogoebel.

PR #50, opened 2017-11-30
Fixes #21821 - Add Rails 5 style migrations

xprazak2:
To be merged after core moves to Rails 5.1


lzap:
The core is now on Rails 5, merging to unblock tests.

PR #52, opened 2017-12-06
Fixes #21877 - fixed rake tasks with taxonomy

lzap:
Taxonomy is now enforced which returns nil for the templates. Both templates are global, thus taxonomy is not relevant. For rake tasks, this is also irrelevant. So added `unscoped` everywhere.


lzap:
The patch does not work for host bootdisks, I will need to add as-admin.

lzap:
Ok now it works.

timogoebel:
Works well, thanks @lzap. I'll wait for ci tests and then merge this.

PR #53, opened 2017-12-28
Foreman 1.17 updates

mmoll:


mmoll:
Somebody with actual knowledge of the plugin should have a look at the `fallback_location:` parameters, but at least I'm getting green tests locally with this.

ares:
Makes sense to me, the disks controller does not have a good default action so '/' sounds good, it's only used when there's some unexpected error and there's no referer which should happen very rarely. @timogoebel could you please take a look to unblock tests? @lzap is afk until next week.

timogoebel:
Merged, thanks @mmoll and @ares.

PR #55, opened 2018-01-10
Refs #14797 - disable media presence check

lzap:
This is a workaround for 10.x branch until we refactor this in core.


lzap:
Tests are broken, I prefer downstream only patch in this case.

PR #57, opened 2018-02-01
Fixes #22482 - bootdisk_raise works in host context

lzap:
Did not work in host context, only in controller context.

### DIFF
--- a/app/services/foreman_bootdisk/renderer_methods.rb
+++ b/app/services/foreman_bootdisk/renderer_methods.rb
@@ -10,14 +10,18 @@ module ForemanBootdisk
         u.fragment = nil
         u.to_s
       end
+
+      def bootdisk_raise(*args)
+        raise ::Foreman::Exception.new(*args)
+      end
     end
 
-    def bootdisk_chain_url(mac = '', action = 'iPXE')
+    def bootdisk_chain_url(mac = self.try(:mac), action = 'iPXE')
       self.class.format_bootdisk_chain_url(foreman_url(action), mac)
     end
 
     def bootdisk_raise(*args)
-      raise ::Foreman::Exception.new(*args)
+      self.class.bootdisk_raise(*args)
     end
   end
 end
--- a/lib/foreman_bootdisk/engine.rb
+++ b/lib/foreman_bootdisk/engine.rb
@@ -77,6 +77,7 @@ module ForemanBootdisk
     config.to_prepare do
       begin
         Host::Managed.send(:prepend, ForemanBootdisk::HostExt)
+        Host::Managed.send(:include, ForemanBootdisk::RendererMethods)
         Host::Managed.send(:include, ForemanBootdisk::Orchestration::Compute) if SETTINGS[:unattended]
         HostsHelper.send(:prepend, ForemanBootdisk::HostsHelperExt)
         UnattendedController.send(:prepend, ForemanBootdisk::UnattendedControllerExt)
--- a/app/models/concerns/foreman_bootdisk/host_ext.rb
+++ b/app/models/concerns/foreman_bootdisk/host_ext.rb
@@ -10,10 +10,6 @@ module ForemanBootdisk::HostExt
     unattended_render(bootdisk_template.template)
   end
 
-  def bootdisk_chain_url(mac = self.mac, action = 'iPXE')
-    ForemanBootdisk::Renderer.format_bootdisk_chain_url(foreman_url(action), mac)
-  end
-
   def bootdisk_build?
     provision_method == 'bootdisk'
   end


lib/foreman_bootdisk/engine.rb
```diff
@@ -77,6 +77,8 @@ class Engine < ::Rails::Engine
     config.to_prepare do
       begin
         Host::Managed.send(:prepend, ForemanBootdisk::HostExt)
+        Host::Managed.send(:extend, ForemanBootdisk::RendererMethods)
```
timogoebel: Why do we need `extend` here?

lib/foreman_bootdisk/engine.rb
```diff
@@ -77,6 +77,8 @@ class Engine < ::Rails::Engine
...
+        Host::Managed.send(:include, ForemanBootdisk::RendererMethods)
```
timogoebel: Shouldn't this be covered by [this](https://github.com/theforeman/foreman/blob/91c553d00ad497df40ac5bdd28945b5f29f34020/app/models/host/managed.rb#L168) and [this](https://github.com/theforeman/foreman_bootdisk/blob/master/lib/foreman_bootdisk/engine.rb#L67)?

lib/foreman_bootdisk/engine.rb
lzap: You are right, extend is not needed.

lib/foreman_bootdisk/engine.rb
lzap: This is actually the fix. The method was made visible only in controller context (`extend_template_helpers`), not in host context. There is no plugin API for that, it is perhaps assumed that plugins include their own concern into Host class.

This needs overhaus as I said, extending Host class like that is messy, Host is God-object in the current design. But that's out of table for now :)

lzap:
Removed the extra line there.

lib/foreman_bootdisk/engine.rb
timogoebel: Actually, Host is the `Oh, my god!` object. :-D

app/services/foreman_bootdisk/renderer_methods.rb
```diff
@@ -10,14 +10,18 @@ def format_bootdisk_chain_url(url, mac)
...
     end
 
-    def bootdisk_chain_url(mac = '', action = 'iPXE')
+    def bootdisk_chain_url(mac = self.try(:mac), action = 'iPXE')
       self.class.format_bootdisk_chain_url(foreman_url(action), mac)
     end
 
     def bootdisk_raise(*args)
-      raise ::Foreman::Exception.new(*args)
+      self.class.bootdisk_raise(*args)
```
timogoebel: Is this really required? Where is `Host.bootdisk_raise` called?

lzap: It's not really, I did it for consistency. Removing.

lzap: Oh wait it is needed, it's controller class vs instance. We need both. When testing, try it without this.

timogoebel:
[test]

JOJ0:
Hi,
I don't know if this is of any help but I tried this patch and still get this error:

2018-04-05T18:29:36 38338656 [templates] [I] Rendering template 'Unnamed'
2018-04-05T18:29:36 38338656 [app] [W] Failed to render boot disk template
 | NoMethodError: undefined method `bootdisk_raise' for #<Host::Managed:0x000000000dc7db40>
 | /usr/share/foreman/vendor/ruby/2.3.0/gems/activemodel-5.1.4/lib/active_model/attribute_methods.rb:432:in `method_missing'
 | /usr/share/foreman/vendor/ruby/2.3.0/gems/safemode-1.3.2/lib/safemode/scope.rb:37:in `method_missing'
 | Unnamed:58:in `bind'
 | /usr/share/foreman/vendor/ruby/2.3.0/gems/safemode-1.3.2/lib/safemode.rb:51:in `eval'
 | /usr/share/foreman/vendor/ruby/2.3.0/gems/safemode-1.3.2/lib/safemode.rb:51:in `eval'


 Is there other pulls/commits I would have to apply to get this working?  I just patched my files by hand with the code in the pull requests commits. I have 1.17 RC2 deb packages installed on Ubuntu Xenial. 

Thanks for any hints.
all the best!
Jojo

lzap:
Hey, Foreman version? I think this might only work with 1.18+ (develop version). You need this patch: http://projects.theforeman.org/issues/22411

JOJ0:
@lzap Sorry I wasn't aware hat there is another version number for foreman other than 1.16 1.17-RCX or the like. I still can't find it. 

this is what foreman-debug gives me:

```
  OS: debian
  RELEASE: stretch/sid
  FOREMAN: 1.17.0-RC2
     RUBY: ruby 2.3.6p384 (2017-12-14 revision 61254) [x86_64-linux-gnu]
   PUPPET: 5.5.0
```

Anyway, thanks for the patch you pointed me to. I tried putting it into place but still no luck, same error. Probably another dependency. If I had the time and resources I would build a 1.18 from directly from github but unfortunately I have to move on and for now will continue my foreman efforts another time :-/

One last question: Both of these patches are pointed to release 1.18, so there is no chance that they will be backported to the final 1.17 release? 

Thanks again!
Jojo

lzap:
Well 1.18 is unreleased version, develop branch actually. We won't backport this as there is an easy fix - just edit your template and comment out the raise for now. You better wait until this PR gets through review and is merged in - it is possible it still does not work and we will figure it out.

JOJ0:
Hi, not sure if I did it correctly. I will try to find time to double check. Yes I restarted apache. 

To clarify things: The error message mentionend abvore is triggered in the GUI and in the log. It happens when I click on an already existing host and click on "Bootdisk - Host '<hostname>' image"

One thing that I discovered just now by accident: I have one host that does not show this error and I can successfully download an iso image.

Sorry I am a Foreman Newbie. In which template do I have to look for the "raise" I searched for the term raise in all my _used_ default, finish, ipxe and so on template but couldn't find this term. In what type of template would I find this "raise" statement would be a starter that could help me.

lzap:
Search for "bootdisk" templates in Provisioning templates, you will find two. One is used for generic hosts, other for host based disks. For full host image PXELinux is used, that's irrelevant in this case. Remove those raise_bootdisk lines and save, then regenerate your image.

Note that the raise is used to error out with a message, it is possible that your host is missing something (subnet, IP address or whatever). Then your bootdisk might not be valid.

ares:
Could someone with merge permissions get this in? Seems like it was reviewed.

timogoebel:
Thanks, @lzap.

PR #62, opened 2018-05-30
Fixes #4413 - Full-host bootdisk image UEFI compatible

lzap:
I want to start over, rebased. This is WIP.

By default this patch should not change anything, but if user copies
efiboot.img and grubx64.efi into a directory and sets this in the settings,
Full Host image will be created with EFI in mind (as hybrid ISO as well). So
the ISO will work both on BIOS and EFI systems.

For RHEL systems, the file can be only find in the kickstart tree:

http://ftp.fi.muni.cz/pub/linux/centos/7/os/x86_64/images/

Grub2 file can be found in grub2-efi package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,5 @@ locale/*/*.pox
 tags
 TAGS
 .foreman_app
-bootloaders
+/bootloaders
+/foreman_bootdisk*gem
--- a/app/models/setting/bootdisk.rb
+++ b/app/models/setting/bootdisk.rb
@@ -1,4 +1,6 @@
 class Setting::Bootdisk< ::Setting
+  BLANK_ATTRS << "bootdisk_efiboot_dir"
+
   def self.load_defaults
     return unless ApplicationRecord.connection.table_exists?('settings')
     return unless super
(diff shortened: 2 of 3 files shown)


dforsberg:
Hi,

Found a issue with this. You cant EFI boot RHEL 6 and alike with Grub2. You get: error kernel to old. We need Grub1 to boot those.

All this wouldn't be a issue if Red Hat were to provide the full iso images in there kickstart/os mirrors. They have started to strip them down to only include the absolutely needed from 6.8/7.0. They dont provide the full iso tree in Foreman like CentOS for example.

If all was there, or if they at least re-added the ../images/efiboot.img everywhere again. Then we could just grab that if it exists and add it to the full iso we create here. Just like vmlinuz and initrd is added now. This could also be done for the isolinux files.
This is what most other distros have in there isos as well. It would get kind of os independent.

iPXE EFI is broken att the moment so that we cant use right now. There is no way to pass custom input to it as of today.

lzap:
Oh yeah I completely forgot about RHEL6, shall we work together on a solution?  I can implement your idea above and copy grub1 too, that should work.

You can download grub1 from here: https://github.com/theforeman/foreman-bootloaders these are from CentOS and Fedora.

PR #64, opened 2018-06-21
Fixes #4413 - Bootdisk images UEFI compatible

dforsberg:
This is a new version of @lzap PR #62 

This uses the grub2 files for foreman-bootloaders-tftpboot located in /var/lib/tftpboot/grub2. 
Supports EFI boot of older operating systems like RHEL6, use _linux_ and _initrd_ command in the PXEGrub2 template for those.

If there is no PXEGrub2 template assigned to the operating system it will only give a warning in the log file. PXEGrub templates are not assigned to operating systems by default and is not required.

TODO: 
Patch for ipxe.efi to read cmdline options: https://bugzilla.redhat.com/show_bug.cgi?id=1401491
Needs a selinux change to work:
```
module satellite-efi-bootdisk 1.0;

require {
        type passenger_t;
        type tftpdir_rw_t;
        class file { getattr ioctl open read };
        class dir search;
}

#============= passenger_t ==============
allow passenger_t tftpdir_rw_t:file { getattr ioctl open read };
allow passenger_t tftpdir_rw_t:dir search;
```
Compile this to a selinux package and install it:
```
checkmodule -M -m -o satellite-efi-bootdisk.mod satellite-efi-bootdisk.te
semodule_package -o satellite-efi-bootdisk.pp -m satellite-efi-bootdisk.mod
semodule -i satellite-efi-bootdisk.pp
```

### DIFF
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -10,10 +10,15 @@ class ForemanBootdisk::ISOGenerator
   def self.generate_full_host(host, opts = {}, &block)
     raise ::Foreman::Exception.new(N_('Host is not in build mode, so the template cannot be rendered')) unless host.build?
 
-    tmpl = host.send(:generate_pxe_template, :PXELinux)
-    unless tmpl
+    pxelinux_template = host.send(:generate_pxe_template, :PXELinux)
+    unless pxelinux_template
       err = host.errors.full_messages.to_sentence
       raise ::Foreman::Exception.new(N_('Unable to generate disk template, PXELinux template not found or: %s'), err)
+    end
+
+    pxegrub_template = host.send(:generate_pxe_template, :PXEGrub2)
+    unless pxegrub_template
+        ForemanBootdisk.logger.warn("Unable to generate disk template, PXEGrub2 template not found. Bootdisk will not support UEFI")
     end
 
     # pxe_files and filename conversion is utterly bizarre
@@ -24,13 +29,14 @@ class ForemanBootdisk::ISOGenerator
       bootfile_info.map do |f|
         suffix = f[1].split('/').last
         iso_f0 = iso9660_filename(f[0].to_s + '_' + suffix)
-        tmpl.gsub!(f[0].to_s + '-' + suffix, iso_f0)
+        pxelinux_template.gsub!(f[0].to_s + '-' + suffix, iso_f0)
+        pxegrub_template.gsub!(f[0].to_s + '-' + suffix, '/' + iso_f0) if pxegrub_template
         ForemanBootdisk.logger.debug("Boot file #{iso_f0}, source #{f[1]}")
         [iso_f0, f[1]]
       end
     end
 
-    generate(opts.merge(:isolinux => tmpl, :files => files), &block)
+    generate(opts.merge(:isolinux => pxelinux_template, :grub => pxegrub_template, :files => files), &block)
   end
 
   def self.generate(opts = {}, &block)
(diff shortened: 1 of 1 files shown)


app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -73,17 +93,35 @@ def self.generate(opts = {}, &block)
         end if opts[:files].respond_to? :each
       end
 
+      if opts[:grub]
+        FileUtils.mkdir_p(File.join(wd, 'build', 'EFI', 'BOOT'))
+        FileUtils.cp('/var/lib/tftpboot/grub2/shimx64.efi', File.join(wd, 'build', 'EFI', 'BOOT', 'BOOTX64.efi'))
```
lzap: We cannot use this path directly, it won't be present on Foreman when Smart Proxy TFTP feature was not installed. Can you add new `Setting[:bootdisk_grub2_dir]` which will default to `/boot/efi/EFI/*` and use that here and on the other three lines? You will need to expand shell glob because the path is different on different platforms: /boot/efi/EFI/fedora vs /boot/efi/EFI/redhat vs /boot/efi/EFI/centos etc.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -41,6 +47,15 @@ def self.generate(opts = {}, &block)
       initrd /script
     EOS
 
+    opts[:grub] = <<-EOS if opts[:grub].nil? && opts[:ipxe]
+      set default=0
+      set timeout=5
+      menuentry "Chainload iPXE chain" {
+        search --no-floppy --set=root -f /ipxe.efi
+        chainloader ($root)/ipxe.efi cmdline "echo If patch was applied you should see this text ; sleep 10"
```
lzap: Change this so it actually works, we need to error out early for static IP allocation and make generic bootdisk working. Pseudo code:

```ruby
if opts[:grub].nil? && opts[:ipxe]
  raise ::Foreman::Exception.new(N_('Cannot generate UEFI iPXE disk for DHCP-less network')) unless host.provision_interface.subnet.dhcp_boot_mode?
  iface = host.host_param("bootdisk_if") || "net0"
  opts[:grub] = <<-EOS
    ...
    chainloader ($root)/ipxe.efi cmdline "dhcp #{iface} & chain #{bootdisk_chain_url}${#{iface}/mac}"
```

You get the idea.

app/services/foreman_bootdisk/iso_generator.rb
dforsberg: unfortunately the grubx64.efi provided in that directory doesn't include the linux and initrd commands... needs change of grub2 rpm in os to work.

dforsberg: from grub2.spec in RHEL7

```
GRUB_MODULES="  all_video boot btrfs cat chain configfile echo efifwsetup \
                efinet ext2 fat font gfxmenu gfxterm gzio halt hfsplus iso9660 \
                jpeg linuxefi lvm minicmd normal part_apple part_msdos \
                part_gpt password_pbkdf2 png reboot search search_fs_uuid \
                search_fs_file search_label sleep test video xfs \
                mdraid09 mdraid1x blscfg multiboot2 multiboot tftp"
```

from foreman-generate-bootloaders cmd:
```
MODULES="all_video boot btrfs cat configfile echo ext2 fat font gfxmenu gfxterm gzio halt hfsplus iso9660 jpeg loadenv loopback lvm mdraid09 mdraid1x minicmd normal part_apple part_msdos part_gpt password_pbkdf2 png reboot search search_fs_uuid search_fs_file search_label serial sleep syslinuxcfg test tftp video xfs linux"
```

lzap: Good point, can you make a comment in the setting "Install foreman-bootloaders-redhat package or copy shimx64.efi and grubx64.efi manually" so users can figure this out both on CentOS/RHEL and Debian/Ubuntu? We don't generate those on Debians.

lzap: And of course keep using the `/var/lib/tftpboot/grub2` path, that's fine.

lzap:
Hello? Any updates?

lzap:
I am going to resurrect this patch, it looks like in UEFI environment Grub2 can do HTTP just fine, few fixes has been merged into Fedora Rawhide recently so this enables us to remove iPXE from the UEFI workflow completely and just rely on Grub2. I will update the changes accordingly and make a new PR.

PR #66, opened 2018-07-13
Refs #21833 - fixed initialization order for bootsnap

lzap:
I was made aware by @carbonin that this prevents Bootsnap from correctly starting tests when Rails Engine is enabled. I was trying to figure this out for months! https://github.com/Shopify/bootsnap/issues/106

I have no idea why we have this in our codebase, but the endpoint just works after this change. If tests pass let's give it a go!


timogoebel:
Thanks, @lzap.

PR #67, opened 2018-09-02
Fixes #24781 - Supprting Rails 5.2

ik5:
Note: There are few issues with template rendering that are due to the massive rewrite of the rendering engine, and they are not part of this PR scope, and require additional care (cc @timogoebel )


timogoebel:
I‘m on pto this week. I thought I fixed all the rendering engine related issues. Can take a look next week.

PR #68, opened 2018-10-03
fixes #25121 - remove explicit asset precompile

timogoebel:
@lzap: This is not required as of 1.16. We can safely remove the code.
Can you include this before you cut a new release?


lzap:
Oh damn I just released it. Is it big deal? Do I need to do a minor release for that?

timogoebel:
@lzap: Don't worry about it, it was dead code in the latest foreman releases.

PR #69, opened 2018-10-08
fixes #25120 - use medium provider interface

timogoebel:
This commit makes use of the new medium provider interface to build the full host image as the old way of calling the methods is deprecated and will be removed in 1.22.
The embedded PXElinux template is now rendered in it's own template scope (yay, new renderer!) so we can avoid ugly `gsub` string replacement hacks.

This also prepares foreman_bootdisk for frozen strings.

And all of this removes this loop, which is very hard to digest:

https://github.com/theforeman/foreman_bootdisk/blob/15ed6d9a84eb2920abced036691a8ccc2c3f0f13/app/services/foreman_bootdisk/iso_generator.rb#L23-L31


test/unit/iso_generator_test.rb
```diff
@@ -3,46 +3,36 @@
...
-      {:"boot/RedHatEnterpriseLinuxAtomic-7.3-x86_64"=>"/tmp/pxeboot/vmlinuz"},
-      {:"boot/RedHatEnterpriseLinuxAtomic-7.3-x86_64"=>"/tmp/pxeboot/initrd.img"}
-    ]
-    @host.operatingsystem.expects(:pxe_files).returns(boot_files)
-    ForemanBootdisk::ISOGenerator.expects(:generate).with({
-      :isolinux => 'Fake kernel line BOOT/NUXATOMIC_7_3_X86_64_VMLINUZ',
-      :files => [[['BOOT/NUXATOMIC_7_3_X86_64_VMLINUZ', '/tmp/pxeboot/vmlinuz']],
-        [['BOOT/ATOMIC_7_3_X86_64_INITRD_IMG', '/tmp/pxeboot/initrd.img']]]}, anything)
-    ForemanBootdisk::ISOGenerator.generate_full_host(@host)
-  end
```
lzap: This test is actually for `iso9660_filename` to make sure that filenames gets shortened to maximum size for ISO format, can you keep it in some way or another? Note `boot/RedHatEnterpriseLinuxAtomic-7.3-x86_64` -> `BOOT/NUXATOMIC_7_3_X86_64_VMLINUZ` (shortened automatically).

timogoebel: Sure, I can add some tests for `iso9660_filename` method. We should definitely have them.

lzap: Yep should be probably a unit test, whatever you prefer.

PR #75, opened 2019-02-05
Fixes #25992 - pulp redirect for full disk fixed

lzap:
This fixes 302 response from pulp for ondemand sync repos.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,8 +14,8 @@ Style/FormatStringToken:
 Style/DoubleNegation:
   Enabled: false
 
-Metrics/LineLength:
-  Max: 180
+Metrics:
+  Enabled: false
 
 Naming/MethodName:
   Exclude:
@@ -24,24 +24,6 @@ Naming/MethodName:
 Rails/SkipsModelValidations:
   Exclude:
     - 'db/migrate/**/*'
-
-Metrics/ClassLength:
-  Enabled: false
-
-Metrics/AbcSize:
-  Enabled: false
-
-Metrics/MethodLength:
-  Enabled: false
-
-Metrics/CyclomaticComplexity:
-  Enabled: false
-
-Metrics/PerceivedComplexity:
-  Enabled: false
-
-Metrics/BlockLength:
-  Enabled: false
 
 Rails/Date:
   Exclude:
@@ -54,3 +36,12 @@ Style/ClassAndModuleChildren:
 Naming/FileName:
   Exclude:
     - 'db/seeds.d/**/*'
+
+# Support both ruby19 and hash_rockets
+Style/HashSyntax:
+  EnforcedStyle: no_mixed_keys
+  Enabled: true
+
+# Both double and single quotes are OK
+Style/StringLiterals:
+  Enabled: false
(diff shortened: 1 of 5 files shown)


timogoebel:
@lzap: What do you think of switching to a proper file downloading library like https://github.com/janko/down ?

lzap:
I feel like it's probably not necessary, we only have this place and it served well. It's no rocket science I guess.

timogoebel:
@lzap: Where is the code that handles `limit`?

lzap:
Whoops. As I said, this is not a rocket science ;-)

timogoebel:
Looks good, but can we please add some tests for this?

lzap:
Absolutely, good idea. Pulled webmock for that, ideal tool.  Also copied few more Rubocop rules from core which are bugging me.

test/unit/iso_generator_test.rb
```diff
@@ -14,7 +14,20 @@ class IsoGeneratorTest < ActiveSupport::TestCase
       let(:template) { FactoryBot.create(:provisioning_template, template: 'Fake kernel line <%= @kernel %> - <%= @initrd %>') }
 
       setup do
-        host.expects(:provisioning_template).with(kind: :PXELinux).returns(template)
+        host.stubs(:provisioning_template).with(kind: :PXELinux).returns(template)
+      end
+
+      test 'fetch handles redirect' do
+        begin
```
timogoebel: You can instead use `Dir.mktmpdir do |dir|` to create a temporary dir that gets automatically deleted at the end of the block.

```ruby
Dir.mktmpdir do |dir|
  # ...
  ForemanBootdisk::ISOGenerator.fetch(File.join(dir, 'bootdisk-test-redirect'), url)
end

File: .rubocop.yml [hunk 2]
```diff
@@ -54,3 +36,10 @@ Style/ClassAndModuleChildren:
 Naming/FileName:
   Exclude:
     - 'db/seeds.d/**/*'
+# Support both ruby19 and hash_rockets
```
timogoebel: nit: missing newline

.rubocop.yml
```diff
@@ -14,8 +14,8 @@ Style/FormatStringToken:
 Style/DoubleNegation:
   Enabled: false
 
-Metrics/LineLength:
-  Max: 180
+Metrics:
+  Enabled: false
```
timogoebel: ❤️

.rubocop.yml
```diff
@@ -54,3 +36,10 @@ Style/ClassAndModuleChildren:
 Naming/FileName:
   Exclude:
     - 'db/seeds.d/**/*'
+# Support both ruby19 and hash_rockets
+Style/HashSyntax:
+  Enabled: false
```
timogoebel: Can we change this to `EnforcedStyle: no_mixed_keys` so you have to at least decide for one or the other syntax?

lzap:
All done!

.rubocop.yml
timogoebel: @lzap: This is still missing. ;-)

lzap:
Fixed tests hopefully.

timogoebel:
Thanks, @lzap.

PR #74, opened 2019-02-05
Fixes #25986 - generic disk generates with katello

lzap:
New media provider broke behavior of the hack in bootdisk.

PR #77, opened 2019-04-02
Fixes #26512 - stub content_source for generic disk

lzap:
When generic bootdisk is used with Katello, it fails due to missing method which is needed. Although providing a stub in bootdisk for Katello might sound weird and I would be totally against tight coupling of plugins, the whole way of generating generic bootdisk images is a hack and should be rewritten from the ground up.

FYI this is continuation of https://github.com/theforeman/foreman_bootdisk/pull/74 where I forgot to update the other part of the if statement.


timogoebel:
Thanks, @lzap.

lzap:
Danke.

PR #80, opened 2019-06-25
Fixes #27125 - improve host template for IPv6

lzap:
It looks like iPXE does not support static IPv6 setting (DHCPv6/RADV to some degree tho). This implements an error message but prepares for the future.

### DIFF
--- a/app/views/foreman_bootdisk/host.erb
+++ b/app/views/foreman_bootdisk/host.erb
@@ -7,9 +7,23 @@
 <%
 interface = @host.provision_interface
 bootdisk_raise(N_('Host has no provisioning interface defined')) unless interface
-bootdisk_raise(N_('Host has no IP address defined')) if interface.ip.nil? || interface.ip.empty?
-bootdisk_raise(N_('Host has no subnet defined')) unless interface.subnet
+bootdisk_raise(N_('Host has no IPv4 or IPv6 address defined')) unless interface.ip.present? || interface.ip6.present?
+bootdisk_raise(N_('Host has no subnet defined')) unless interface.subnet || interface.subnet6
 bootdisk_raise(N_('Host has no domain defined')) unless interface.domain
+if interface.ip.present? && interface.subnet
+  ip = interface.ip
+  subnet = interface.subnet
+elsif interface.ip6.present? && interface.subnet6
+  ip = interface.ip6
+  subnet = interface.subnet6
+  bootdisk_raise(N_('Host bootdisk does not work with static IPv6'))
+else
+  bootdisk_raise(N_('Both IP and Subnet must be set'))
+end
+
+mask = subnet.mask
+gw = subnet.gateway
+dns = subnet.dns_servers.first
 %>
 
 echo Foreman Bootdisk: Host image (<%= @host.name %>)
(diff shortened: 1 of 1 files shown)


app/views/foreman_bootdisk/host.erb
```diff
@@ -27,17 +41,16 @@ goto loop_success
...
 echo Configuring net${idx} for static IP address <%= interface.ip %>
 ifopen net${idx}
-set netX/ip <%= interface.ip %>
-set netX/netmask <%= interface.subnet.mask %>
-<% if interface.subnet.gateway.present? %>
-set netX/gateway <%= interface.subnet.gateway %>
+set net${idx}/ip <%= ip %>
+set net${idx}/netmask <%= mask %>
+<% if gw.present? %>
+set net${idx}/gateway <%= gw %>
```
lzap: I noticed we actually render `set netX/ip` there which seems incorrect!? Dunno how this could work.

app/views/foreman_bootdisk/host.erb
```diff
@@ -7,9 +7,23 @@
...
+if interface.ip.present? && interface.subnet
+  ip = interface.ip
+  mask = interface.subnet.mask
+  gw = interface.subnet.gateway
+  dns = interface.subnet.dns_servers.first
+elsif interface.ip6.present? && interface.subnet6
+  ip = interface.ip6
+  mask = interface.subnet6.mask
+  gw = interface.subnet6.gateway
+  dns = interface.subnet6.dns_primary || interface.subnet6.dns_secondary
```
timogoebel: You should be able to use `subnet6.dns_servers.first` here as well.

How about
```ruby
if interface.ip.present? && interface.subnet
  subnet = interface.subnet
elsif interface.ip6.present? && interface.subnet6
  subnet = interface.subnet6
end

subnet.mask
subnet.dns_servers.first
subnet.gateway
```

lzap: For some reason I could not find, it would not work. Safemode was the issue, I could not understand it because we allow this method for Subnet and IPv6 is not a subclass, is it?

timogoebel: We have `Subnet::Ipv4` and `Subnet::Ipv6`.

lzap: Oh I think I know - I was testing this out on Foreman 1.20 and it might not have this method in safemode allowed methods.

timogoebel: Then it's not available for both IPv4 and IPv6.

lzap:
Well, it distilled down to this. Rebased.

lzap:
I tested this end to end for a customer, but with safemode turned off. Looks like I need to check safemode.

lzap:
There was a last minute typo, fixed. Tests are failing tho for a different reason: https://community.theforeman.org/t/bootdisk-develop-tests-are-broken/14741

lzap:
I was digging in the iPXE codebase and I found the undocumented variable `netX`. It's actually a special variable which represents the "last opened NIC". We actually use that later in the process (kickstart iPXE template) so I am keeping using that here and I am adding a comment.

lzap:
I am merging this, doing new release and will be testing this over the test week 1.23.

PR #83, opened 2019-08-22
Fixes #27668 - Allow bootdisk_build? in safemode

Rohan21Lobo:
We had merged this previously https://github.com/theforeman/foreman/pull/6989
But then reverted since method should be white listed by the plugin itself.

This change should do the job.


ekohl:
ok to test

tbrisker:
Looks like tests passed but rubocop failed to to linting issues unrelated to this PR

lzap:
Please use `allowed_template_helpers :bootdisk_chain_url, :bootdisk_raise` in `engine.rb` to add new safemode rule.

Also rebase on top of master, I've fixed the Rubocop issue.

PR #85, opened 2019-11-12
Use correct path in `File.exists` check

damonmaria:
Currently `ldlinux.c32` is never copied into the ISO because of this.


timogoebel:
ok to test

Makes sense.

PR #90, opened 2020-03-11
Proxmox extention for foreman_bootdisk

Manisha15:


lib/foreman_bootdisk/engine.rb
```diff
@@ -69,6 +69,7 @@ class Engine < ::Rails::Engine
         Host::Managed.include ForemanBootdisk::Orchestration::Compute if SETTINGS[:unattended]
         HostsHelper.prepend ForemanBootdisk::HostsHelperExt
         Foreman::Model::Vmware.prepend ForemanBootdisk::ComputeResources::Vmware if Foreman::Model::Vmware.available?
+        ForemanFogProxmox::Proxmox.prepend ForemanBootdisk::ComputeResources::Proxmox if ForemanFogProxmox::Proxmox.available?
```
timogoebel: This will fail if `ForemanFogProxmox` is not available.  I believe you can use `defined?` here to check if `ForemanFogProxmox` is defined.

```suggestion
        ForemanFogProxmox::Proxmox.prepend ForemanBootdisk::ComputeResources::Proxmox if defined?(ForemanFogProxmox)
```
But this is untested.

sbernhard:
Thanks Manisha. I guess, it would make sense that @tristanrobert (foreman fog proxmox) should also have a look at this.

Manisha15:
> Thanks for taking the time to add tests.
> However, I saw that the `new_vm`-test only tests if foreman_fog_proxmox does not change the attributes upon calling the `new_server_vm()` (and if it uses server-virtualization).
> This is a pretty small coverage.
> Any chance you could extend this (without mocking a complete proxmox-server  ).
> 
> It would be good to have a test for `iso_upload()` and `iso_attach()` as well.

Thank you for reviewing. I have added tests for iso_upload and iso_attach. Can you please review again.

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

test/unit/concerns/orchestration/compute_test.rb
```diff
@@ -10,13 +10,28 @@ class OrchestrationComputeTest < ActiveSupport::TestCase
...
+      @proxmox_host.send(:setIsoImage)
     end
 
     test 'provisioning a host with provision method bootdisk should upload iso' do
       @cr.expects(:iso_upload)
       @host.send(:setIsoImage)
     end
 
+    test 'provisioning a host with provision method bootdisk in proxmox should attach iso' do
+      @proxmox_cr.expects(:iso_upload)
```
m-bucher: I think you mean `:iso_attach`, here.

test/test_plugin_helper.rb
```diff
@@ -4,6 +4,9 @@
 require 'webmock/minitest'
 require 'webmock'
 
+FactoryBot.definition_file_paths << File.join(File.dirname(__FILE__), 'factories')
```
ekohl, line 7:
```suggestion
FactoryBot.definition_file_paths << File.join(__dir__, 'factories')
```

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

sbernhard:
Looks like the tests are running but some rubocop issues occured in the proxmox extension. Can you have a look at them @Manisha15 ?

See https://ci.theforeman.org/job/foreman_bootdisk-pull-request/36/database=postgresql,label=fast,ruby=2.6/console

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

sbernhard:
Tests and rubocop checks are now successful.

test/test_plugin_helper.rb
```diff
@@ -4,6 +4,9 @@
...
+FactoryBot.reload
+
```
lzap, line 9: Ditto. You can see how VMWare is implemented. There is no difference except VMWare is part of core and Proxmox is part of plugin. If plugin is not found, tests should be skipped.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -0,0 +1,34 @@
...
+      def iso_upload(iso, vm_uuid)
+        server = find_vm_by_uuid(vm_uuid)
+        config_attributes = {
+          'bootdisk' => 'ide2',
+          'boot' => 'dcn'
+        }
+        server.update(config_attributes)
+        server.ssh_options = { password: fog_credentials[:pve_password] }
+        server.ssh_ip_address = bridges.first.address
+        server.scp_upload(iso, '/var/lib/vz/template/iso/')
```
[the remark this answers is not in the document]
Manisha15: I am using CentOS with SELinux in enforcing mode and the provision is working fine on it, so, I guess we don't need to update the SELinux policy?

gemfile.d/foreman_bootdisk.rb
```diff
@@ -0,0 +1,3 @@
+# For testing the foreman bootdisk proxmox extensions
+# this gem file is required
+gem 'foreman_fog_proxmox'
```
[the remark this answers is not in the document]
sbernhard: If you know a better solution to run the tests for foreman_bootdisk together with the proxmox extension tests I'm happy to change this.

test/factories/proxmox_factory.rb
```diff
@@ -0,0 +1,31 @@
...
+               headers: {
+                 'Accept' => 'application/json',
+                 'Host' => '192.168.56.101:8006',
+                 'User-Agent' => 'fog-core/2.1.0'
+               }
+             ).to_return(status: 200, body: "", headers: {})
+    end
+    factory :proxmox_cr, :class => ForemanFogProxmox::Proxmox, :traits => [:proxmox]
+  end
+end
```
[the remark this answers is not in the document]
sbernhard: So you mean, you want to move this change to foreman_fog_proxmox?

gemfile.d/foreman_bootdisk.rb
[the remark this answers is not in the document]
ekohl: AFAIK gemfile.d is only used in tests and is the currently supported way to install additional dependencies. It's not in the gemspec and not used in packaging so I don't really see the problem.

test/test_plugin_helper.rb
sbernhard: I think we are good to go. 
@timogoebel / @lzap / @ekohl ?

sbernhard:
How to proceed with this PR?

gemfile.d/foreman_bootdisk.rb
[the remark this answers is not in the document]
lzap: I haven't realized this is not a new dependency. Although I hate what we are doing here (Proxmox is not the only one), that's ok.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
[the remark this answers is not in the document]
lzap: Haven't realized the path is on target system and not local one. This is good.

test/factories/proxmox_factory.rb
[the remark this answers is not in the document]
lzap: Well I'd expect you have some tests there and FactoryBot could be reused. Not a must tho.

test/test_plugin_helper.rb
lzap: If you are able to reuse factories from the plugin that would be better.

ekohl:
It's something that w do in Jenkins during testing there:
https://github.com/theforeman/foreman-infra/blob/cbe856dcf378c4afef467cf3e0902a4e685c4fc5/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh#L53

Jenkins now fails with:

```
File does not exist: simplecov

rake aborted!
```

That looks unrelated.

I'm not sure what the best way to deal with it locally is, but skipping tests if proxmox is not available does make sense to me.

timogoebel:
> but skipping tests if proxmox is not available does make sense to me.

I believe we do something similar in foreman_dlm.
https://github.com/dm-drogeriemarkt/foreman_dlm/blob/ca43f537d62235969fe64494684f0ca4232e8768/test/models/host_monitoring_test.rb#L9

ekohl:
I'd recommend to at least provide a string why it was skipped, but that's indeed what I was thinking of.

lzap:
> File does not exist: simplecov

I do believe this is related: https://github.com/theforeman/foreman_fog_proxmox/blob/a7d9357817f225a1df9a25a2a043c236a17d9c36/test/test_plugin_helper.rb#L5

I don't see this error locally, please fix this. I am not sure if it's good idea to execute proxmox test in bootdisk plugin jenkins tests - if something changes in proxmox our jenkins will be red. We have the same with core, but plugin-core is kindof well defined, this is plugin-plugin.

At least we could make an environmental option and start executing tests, but if we run into problems we can turn it off.

foreman_bootdisk.gemspec
```diff
@@ -22,7 +22,7 @@ Gem::Specification.new do |s|
     'LICENSE',
     'README.md'
   ]
-  s.files = `git ls-files`.split("\n") - Dir['.*', 'Gem*', '*.gemspec']
+  s.files = `git ls-files`.split("\n") - Dir['.*', 'Gem*', '*.gemspec', 'gemfile.d/*']
```
lzap, line 25: Do we need to distribute this with the gem? This will require packaging changes.

lib/foreman_bootdisk/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module ForemanBootdisk
-  VERSION = '16.1.0'
+  VERSION = '16.0.0'
```
lzap, line 4: Do not touch this line in this PR please. There is a conflict by the way because of this.

gemfile.d/foreman_bootdisk.rb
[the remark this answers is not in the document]
lzap: Honestly, I don't like the `gemfile.d` directory being introduced. I don't see any added value there. Can't we just add this dependency to the main `Gemfile` with something like `if Rails.env.test?`?

foreman_bootdisk.gemspec
sbernhard: As @ekohl said:
AFAIK gemfile.d is only used in tests and is the currently supported way to install additional dependencies. It's not in the gemspec and not used in packaging so I don't really see the problem.

We do this only, to run these tests but don't have a hard dependency for foreman_fog_proxmox in foreman_bootdisk.

sbernhard:
[test foreman_bootdisk]

lib/foreman_bootdisk/version.rb
Manisha15: Fixed !

Manisha15:
This is ready for review.

foreman_bootdisk.gemspec
ekohl: I agree with @lzap here that we don't need to ship it in the gem. We only use it in Jenkins and is really a custom solution that isn't used elsewhere AFAIK. It might just confuse people. What I would instead do is write something about this in README.

gemfile.d/foreman_bootdisk.rb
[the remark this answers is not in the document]
ekohl: No, because Gemfile is simply not used. The relation is inverse: foreman adds a dependency on foreman_bootdisk. That means only the gemspec is used. Then all Rake tasks are executed from Foreman.

test/test_plugin_helper.rb
```diff
@@ -4,6 +4,10 @@
 require 'webmock/minitest'
 require 'webmock'
 
+FactoryBot.definition_file_paths << File.join(File.dirname(__FILE__), 'factories')
+FactoryBot.definition_file_paths << "#{ForemanFogProxmox::Engine.root}/test/factories" if ForemanBootdisk.with_proxmox?
+FactoryBot.reload
```
ekohl, line 9: Out of interest and because I don't know FactoryBot very well: why do you need to reload?

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

test/test_plugin_helper.rb
Manisha15: To load factories from fog_proxmox otherwise it will use only from foreman and raises error in proxmox compute resource related tests that `Factory not registered: "proxmox_cr"`

ekohl: Makes me wonder if Foreman's test helper should just load all factories from all plugins for you. Might be nice to have in core.

sbernhard:
[test foreman_bootdisk]

lzap:
Hello, tests are failing because of Rubocop. Jenkin implementation is a bit werird, this should have been probably a separate job. Hopefuly with GH Actions we can do it properly.

https://ci.theforeman.org/job/foreman_bootdisk-pull-request/76/database=postgresql,label=fast,ruby=2.5/consoleText

sbernhard:
My thought was to use https://github.com/theforeman/foreman/pull/7885 to fix the failing tests. Looks like the referenced PR takes longer as expected.
Is there anything else we can do to get this done?

ekohl:
Fix the rubocop failures to be actually compliant with Foreman's profile? For example, in this repo the `.rubocop.yml` is incompatible with the latest Rubocop so that PR wouldn't even work currently.

I'd suggest to open a clean PR to master just to fix rubocop, then rebase this PR on top of it.

ekohl:
Not sure if I'm reading it right, but it may also be trying to rubocop the installed gems.

sbernhard:
> Fix the rubocop failures to be actually compliant with Foreman's profile? For example, in this repo the `.rubocop.yml` is incompatible with the latest Rubocop so that PR wouldn't even work currently.
> 
> I'd suggest to open a clean PR to master just to fix rubocop, then rebase this PR on top of it.

see https://github.com/theforeman/foreman_bootdisk/pull/97 - hope this is what you have exepected.

sbernhard:
[test foreman_bootdisk]

ekohl:
It looks like the problem is that you're loading multiple plugins. The fog proxmox plugin is also loaded and that task actually fails. It should be possible to locally replicate this by getting a foreman checkout, add the proxmox plugin and run `rake rubocop_foreman_fog_proxmox`. It'd be interesting to see the backtrace.

I'm not sure if you can avoid running Rubocop if the other plugin enhances `jenkins:unit`. You can clear an enhanced task, but that's not very clean.

The whole setup of multiple plugins present is a bit uncharted territory.

lzap:
I still see ton of rubocop issues, all we need to do is to iron them out.

lzap:
For the record, we are having a discussion how to handle rubocop.

https://community.theforeman.org/t/rake-task-to-run-rubocop-on-a-plugin/19979

Here is a PR that implements Rubocop via separate Github Action: https://github.com/theforeman/foreman_bootdisk/pull/98

If it gets merged and we remove rubocop from our own CI, then rebase on top of that.

sbernhard:
Thanks for letting me know. very good idea.

ekohl:
> If it gets merged and we remove rubocop from our own CI, then rebase on top of that.

That won't suffice. Per https://github.com/theforeman/foreman_bootdisk/pull/90#issuecomment-682085906 you need to update proxmox to drop it [here](https://github.com/theforeman/foreman_fog_proxmox/blob/d8094877d9a40e8aaf775553ee4979356156b48b/lib/tasks/foreman_fog_proxmox_tasks.rake#L37).

The problem isn't that **this** plugin calls Rubocop - it's that foreman_fog_proxmox calls Rubocop and doesn't comply to the rules in this repo.

tristanrobert:
> > If it gets merged and we remove rubocop from our own CI, then rebase on top of that.
> 
> That won't suffice. Per https://github.com/theforeman/foreman_bootdisk/pull/90#issuecomment-682085906 you need to update proxmox to drop it [here](https://github.com/theforeman/foreman_fog_proxmox/blob/d8094877d9a40e8aaf775553ee4979356156b48b/lib/tasks/foreman_fog_proxmox_tasks.rake#L37).
> 
> The problem isn't that **this** plugin calls Rubocop - it's that foreman_fog_proxmox calls Rubocop and doesn't comply to the rules in this repo.

@ekohl foreman_fog_proxmox imports Foreman rubocop rules. Where it does not comply? I can fix it.

lzap:
> I always check `bundle exec bin/rake foreman_fog_proxmox:rubocop`

The discussion is going towards dropping this "centralized" rubocop testing via core Rakefile and use a different approach. If you want to unblock this PR, just get rid of the "enhance" hack in your plugin and run Rubocop in a different way (e.g. via Travis, GHA or from your plugin Rakefile):

https://github.com/theforeman/foreman_fog_proxmox/blob/master/lib/tasks/foreman_fog_proxmox_tasks.rake#L37

tristanrobert:
> > I always check `bundle exec bin/rake foreman_fog_proxmox:rubocop`
> 
> The discussion is going towards dropping this "centralized" rubocop testing via core Rakefile and use a different approach. If you want to unblock this PR, just get rid of the "enhance" hack in your plugin and run Rubocop in a different way (e.g. via Travis, GHA or from your plugin Rakefile):
> 
> https://github.com/theforeman/foreman_fog_proxmox/blob/master/lib/tasks/foreman_fog_proxmox_tasks.rake#L37
> 
> Watch the discussion as @ezr-ondrej is working on improving this, creating some shared Rubocop configuration via a rubygem so you can utilize this if you want.

Fixed in https://github.com/theforeman/foreman_fog_proxmox/pull/170

sbernhard:
[test foreman_bootdisk]

lzap:
I still see those cops failing...

It is shaping up, we have a new gem and once it's finished I will proceed with a PR for this plugin: https://community.theforeman.org/t/rubocop-common-styles-in-gem/20794

sbernhard:
whats the state of foreman-rubocop @lzap ? Can we somehow proceed with this PR?

lzap:
Thanks for reminder, it is now my priority to review and merge the changes into core.

sbernhard:
Friendly reminder @lzap :-)

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

lzap:
Rubocop is fine now, please fix the test and we are good to go. This is not pretty, but VMWare is not pretty as well. In general, it would be great if we could refactor this someday - ISO upload/attaching could easily be part of core.

sbernhard:
[test foreman_bootdisk]

sbernhard:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

lzap:
Still many failed tests.

I still have a bit of concern that now when something changes in the Proxmox plugin this can easily get Bootdisk tests red. But I count on you guys that you will act, perhaps quicker than me :-) Also if you can run tests in your Proxmox plugin repo with bootdisk tests that would be awesome.

Manisha15:
@lzap Is there any alternative so as to make these two plugins independent of each other or at least making the tests independent of each other so that change in Proxmox plugin don't fail bootdisk plugin tests?

Manisha15:
[test foreman_bootdisk]

lzap:
I don't think so, and I am fine as long as you could run bootdisk tests also on proxmox changes for each PR before they are actually merged. And feel free to do such improvement in this repo too (so no change would be ever merged if this would break proxmox).

There are failures and it needs a rebase :-(

Manisha15:
[test foreman_bootdisk]

lzap:
To let you know, I am monitoring this PR and once you get tests passing AND you guys update the proxmox CI in a way that you always run bootdisk tests on each PR as well to prevent breaking it. Then we are good to go :-)

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

lzap:
What is the status? How can I help?

Manisha15:
> What is the status? How can I help?

The error is related to access_permission in foreman_fog_proxmox which is used by this plugin:

`permission for foreman_fog_proxmox/compute_resources/ostemplates_by_id_and_node_and_storage not found, check access_permissions.rb`

The permissions are given here:
https://github.com/theforeman/foreman_fog_proxmox/blob/91ca57f45944b5ce503c4276a7859374b41d671f/lib/foreman_fog_proxmox/engine.rb#L49

Do you think I need to give permission to access these 7 components in foreman or in foreman_fog_proxmox ?

lzap:
The plugin adds this action, the plugin needs permissions.

https://github.com/theforeman/foreman_fog_proxmox/blob/ca9475c15731e866cc3d8b1998320b30838f16d1/app/controllers/foreman_fog_proxmox/compute_resources_controller.rb#L31

Look at other plugins for a reference, should be pretty much copy and paste.

Manisha15:
> The plugin adds this action, the plugin needs permissions.
> 
> https://github.com/theforeman/foreman_fog_proxmox/blob/ca9475c15731e866cc3d8b1998320b30838f16d1/app/controllers/foreman_fog_proxmox/compute_resources_controller.rb#L31
> 
> Look at other plugins for a reference, should be pretty much copy and paste.

The access permissions are already added in foreman_fog_proxmox plugin but access_permissions seeds are missing for the tests https://github.com/theforeman/foreman/blob/develop/db/seeds.d/020-permissions_list.rb. How do I add these?

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
[test foreman_bootdisk]

Manisha15:
@lzap The tests are green. Can you please review it again?

gemfile.d/foreman_bootdisk.rb
```diff
@@ -0,0 +1,6 @@
+# this gem is required by foreman_fog_proxmox
+gem 'simplecov'
+
+# For testing the foreman bootdisk proxmox extensions
+# this gem file is required
+gem 'foreman_fog_proxmox', '>= 0.13.4'
```
lzap, line 6: Can you name this file `proxmox.rb` or something similar? So it is clear in the packaging phase what does this carries. Thanks for extracting this into separate file, in Satellite we would not distribute this one. By the way, I do not see this file included from the Gemfile, I think you might want to add a code that includes it.

@stejskalleos this file must not be distributed in Satellite package as Proxmox is not supported or shipped in Satellite repos, make sure there is a BZ prior merging (or preferably a patch)

@theforeman/packaging is this an acceptable solution? do we have a SPEC that would tell us if a package is built for upstream or downstream?

test/test_plugin_helper.rb
```diff
@@ -1,9 +1,16 @@
 # frozen_string_literal: true
 
+#require 'simplecov'
+#SimpleCov.start
+
```
lzap, line 5: A leftover.

test/test_plugin_helper.rb
lzap: I am not so sure about this one as well, but I am not a FactoryBot expert. This could bite tho.

test/unit/concerns/orchestration/compute_test.rb
```diff
@@ -52,6 +106,21 @@ class OrchestrationComputeTest < ActiveSupport::TestCase
...
+      tasks = @proxmox_host.queue.all.map(&:name)
+      assert_not_includes tasks, "Generating ISO image for #{@proxmox_host.name}"
+      assert_not_includes tasks, "Upload ISO image to datastore for #{@proxmox_host.name}"
+      assert_not_includes tasks, "Attach ISO image to CDROM drive for #{@proxmox_host.name}"
+      assert_includes tasks, "Detach ISO image from CDROM drive for #{@proxmox_host.name}"
+    end
+
     test 'the iso should be detached when the host leaves build mode' do
       @host.stubs(:compute?).returns(true)
       old = stub()
```
lzap, line 126: It would be probalbly much better to introduce a new test that only includes those new methods, but I don't insist.

lzap (changes requested):
I am no longer a maintainer of bootdiskplugin, so while I can keep reviewing, merging will be up to someone else. Perhaps @stejskalleos !?

gemfile.d/foreman_bootdisk.rb
evgeni: Of course not. That would make it even more complicated than it already is.

If that's a test only dependency (aka: otherwise the plugin dymamically detects the presence, or absence of f_f_proxmox), it should not be in this file at all. 

We don't want to force upstream users of bootdisk to install proxmox, don't we?

lzap: Man I completely forgot about poor upstream users. You are right, this enforces them to install proxmox, thus making a packaging dependency.

I am not sure how to solve this plugin to plugin dependency, do we have a precedent maybe?

Manisha15: Yes this is test dependency. The tests were failing if I don't add this dependency. Do you any better way to handle it?

lzap: One idea I have is to gate this behind an ENV variable, by default the dependency would not be required thus making all those tests to be skipped. So user who is not interested in testing Proxmox would just not notice it. But in your workflow, you can set the variable and test everything.

evgeni:
```
group :test do
  gem 'blah'
end
```

wouldn't that work?

Manisha15: This works.

evgeni: Nah, this skipped all proxmox tests.

test/unit/concerns/orchestration/compute_test.rb
Manisha15: Do you mean to add only `iso_upload`, `iso_attach` and, `iso_detach` tests and remove others?

gemfile.d/foreman_bootdisk.rb
Manisha15: Yeah, it needs different approach.
I didn't quite understand ENV variable method. @lzap Can you please explain it?

gemfile.d/foreman_bootdisk.rb
```diff
@@ -0,0 +1,5 @@
+# For testing the foreman bootdisk proxmox extensions
+# this gem file is required
+group :test do
```
evgeni, line 3: I think this file is just not loaded at all. Can you try puting the same code into `Gemfile`?

lzap: Yeah I had the same comment above.

test/unit/concerns/orchestration/compute_test.rb
lzap: What I mean was extracting all new methods into a separate TestClass.

gemfile.d/foreman_bootdisk.rb
Manisha15: It still skipped the tests.

evgeni: Right. Because it's stupid.

Can you try adding it here?
https://github.com/theforeman/foreman_bootdisk/blob/913d42b3c9bf032ac67cb7dde4c8230812b0b481/.github/workflows/ruby_ci.yaml#L52
so a second line, a la
```
           echo "gem 'foreman_fog_proxmox'" >> bundler.d/foreman_bootdisk.local.rb 
```

evgeni:
The tests seem to be executed now, but I don't really see why they fail.

Manisha15:
> The tests seem to be executed now, but I don't really see why they fail.

This works and now tests fails with `AccessPermissionsTest` error. It throws error because it doesn't have  access permissions for https://github.com/theforeman/foreman_fog_proxmox/blob/91ca57f45944b5ce503c4276a7859374b41d671f/lib/foreman_fog_proxmox/engine.rb#L49.

gemfile.d/foreman_bootdisk.rb
lzap: Well, now that I am thinking about it, bundler should pull the dependency automatically and since I haven't realized this is a test only dependency, I think the solution suggested by Evgeni is okay. So don't introduce any variables.

Manisha15:
[test foreman_bootdisk]

lib/foreman_bootdisk/engine.rb
```diff
@@ -154,4 +155,11 @@ class Engine < ::Rails::Engine
   def self.logger
     Foreman::Logging.logger('foreman_bootdisk')
   end
+
+  def self.with_proxmox?
+    ForemanFogProxmox # rubocop:disable Lint/Void
+    true
+  rescue StandardError
```
m-bucher, line 162: I would expect this to be a `NameError`.

Alternatively: `Object.const_defined?('ForemanFogProxmox')` might work as well

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -0,0 +1,50 @@
...
+        server.reload
+        storage = storages(server.node_id, 'iso')[0]
+        storage.volumes.any? { |v| v.volid.include? File.basename(iso) }
+      end
+
+      def iso_attach(iso, vm_uuid)
+        server = find_vm_by_uuid(vm_uuid)
+        storage = storages(server.node_id, 'iso')[0]
+        volume = storage.volumes.detect { |v| v.volid.include? File.basename(iso) }
+        disks = server.disks.map {|disk| disk.split(":")[0] }.join(";")
```
m-bucher, line 25:
```suggestion
        disks = server.disks.map { |disk| disk.split(":")[0] }.join(";")
```
Codestyle nitpick :grin: I would expect rubocop to complain about this, but it is configured here to basically check nothing :monocle_face: 
Regardless, I guess it makes sense to have consistent formatting :wink:

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -0,0 +1,50 @@
...
+        config_hash = { ide2: "#{volume.volid},media=cdrom" }
+        server.update(config_hash)
+        server.update({:boot => "order=ide2;#{disks}"})
```
m-bucher, line 28:
```suggestion
        server.update({ boot: "order=ide2;#{disks}" })
```

m-bucher (approved):
Looks good overall. Some minor code-style comments :grin:

lib/foreman_bootdisk/engine.rb
```diff
@@ -154,4 +155,8 @@ class Engine < ::Rails::Engine
   def self.logger
     Foreman::Logging.logger('foreman_bootdisk')
   end
+
+  def self.with_proxmox?
+    Object.const_defined?('ForemanFogProxmox')
```
evgeni, line 160:
```suggestion
    Foreman::Plugin.installed?('foreman_fog_proxmox')
```

Manisha15: updated

Manisha15:
@evgeni @lzap Can you please review this again? The tests are green now.

.github/workflows/ruby_ci.yaml
```diff
@@ -45,6 +45,7 @@ jobs:
       - name: Setup Bundler
         run: |
           echo "gem 'foreman_bootdisk', path: './foreman_bootdisk'" > bundler.d/foreman_bootdisk.local.rb
+          echo "gem 'foreman_fog_proxmox'" >> bundler.d/foreman_bootdisk.local.rb
```
evgeni, line 48: I would have expected this is not required, given you also edited `Gemfile`? (not blocking, just curious)

evgeni:
I am neither a bootdisk nor a proxmox expert, but this doesn't read crazy to me :)

evgeni:
fixing the RPM build in https://github.com/theforeman/foreman_bootdisk/pull/149

.github/workflows/ruby_ci.yaml
Manisha15: Yeah, now it is not required.

evgeni:
Now it skipped the Proxmox tests 😞

Sorry for the bad idea

Manisha15:
> Now it skipped the Proxmox tests 😞
> 
> Sorry for the bad idea

changed it again, and now tests are running and green.

.github/workflows/ruby_ci.yaml
ekohl: FYI: in our common CI we use `gemfile.d/*.rb`. For our reusable GHA it's [here](https://github.com/theforeman/actions/blob/a386c3ff5ce8e9a5faeefc75be67586eb03f9f3e/.github/workflows/foreman_plugin.yml#L75-L80) and Jenkins implements a similar thing.

Then in `Gemfile` you can use `eval_gemfile`:
```ruby
group :test do
  eval_gemfile 'gemfile.d/proxmox.rb'
end
```

This means you only specify `>= 0.14.2` once.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      def iso_attach(iso, vm_uuid)
+        server = find_vm_by_uuid(vm_uuid)
+        storage = storages(server.node_id, 'iso')[0]
+        volume = storage.volumes.detect { |v| v.volid.include? File.basename(iso) }
+        disks = server.disks.map { |disk| disk.split(":")[0] }.join(";")
+        config_hash = { ide2: "#{volume.volid},media=cdrom" }
+        server.update(config_hash)
+        server.update({ boot: "order=ide2;#{disks}" })
```
ekohl, line 28: AFAIK the latest rubocop-performance will flag this and tell you to simplify to regular assignments:
```suggestion
        server[:ide2] = "#{volume.volid},media=cdrom"
        server[:boot] = "order=ide2;#{disks}"
```

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -0,0 +1,50 @@
...
+      def iso_detach(vm_uuid)
+        server = find_vm_by_uuid(vm_uuid)
+
+        # get volid to delete iso after detaching from vm
+        volid = server.volumes.get("ide2").volid
+        config_hash = { ide2: "none,media=cdrom" }
+        server.update(config_hash)
+
+        # cdrom will be ejected on next power off
+        server.detach('ide2')
```
ekohl, line 41: Given the repetition, would it be good to declare `ide2` as a constant?

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        # delete the iso file from proxmox server
+        storage = storages(server.node_id, 'iso')[0]
+        volume = storage.volumes.detect { |v| v.volid.include? volid }
```
ekohl, line 45: When would this volume ID not be exactly equal? If the volume IDs are integers I'd imagine this also end up being `123'.include?('12')` so
```suggestion
        volume = storage.volumes.detect { |v| v.volid == volid }
```

Also, previously you already used `volid = server.volumes.get("ide2").volid` so why can't you just do:
```
server.volumes.get('ide2').destroy
```

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -0,0 +1,50 @@
...
+        server.reboot
+      end
+
+      def iso_detach(vm_uuid)
+        server = find_vm_by_uuid(vm_uuid)
+
+        # get volid to delete iso after detaching from vm
+        volid = server.volumes.get("ide2").volid
+        config_hash = { ide2: "none,media=cdrom" }
+        server.update(config_hash)
```
ekohl, line 38:
```suggestion
        server[:ide2] = "none,media=cdrom"
```

ekohl (changes requested):

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
Manisha15: We don't want to destroy the disk 'ide2' as the operation is not possible when vm is powered on. We just want to destroy the volume so that the `iso files` are not piled up in proxmox server.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
Manisha15: Since it's a server object and not a hash so this assignment is not possible and it throw error `NoMethodError (undefined method `[]=' for #<Fog::Proxmox::Compute::Server:0x0000564003398f20>)`

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
Manisha15: Since it's a server object and not a hash so this assignment is not possible and it throw error NoMethodError (undefined method []=' for #Fog::Proxmox::Compute::Server:0x0000564003398f20)`

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
Manisha15: ide2 is a cdrom and for attachment and detachment it has different values. In iso detach function, first it is ejected, thus, set to none and  detach action takes place which will be executed when the vm is powered off. So, it cannot be set as a constant.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
Manisha15: And we want to first add the cdrom (ide2) and update the server and then update the boot order because if we update the server with both cdrom and boot order, it takes wrong boot order. It tries to update the attributes alphabetically and without adding `ide2` it cannot update boot order properly.

.github/workflows/ruby_ci.yaml
Manisha15: @ekohl This also fails the rubocop as it cannot find the `gemfile.d/proxmox.rb`.

ekohl: > Added this change. But how will it find the `gemfile.d/proxmox.rb` as it's not there on foreman side as well as in foreman_bootdisk `gemfile.d` ?

Foreman doesn't care about `Gemfile`, so it shouldn't matter.

> @ekohl This also fails the rubocop as it cannot find the `gemfile.d/proxmox.rb`.

Because you didn't use `git add` so it's not in the repository.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
ekohl: But it's hardcoded in the code, so it already is. I don't see why you can't use `CDROM_VOLUME = 'ide2'` and then later `server.detach(CDROM_VOLUME)`.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
ekohl: So `volume = storage.volumes.detect { |v| v.volid.include? volid }` returns a different object than `server.volumes.get("ide2")`. That would surprise me.

Manisha15: Yes, server.volumes are disks of the server so it returns the disks, `Fog::Proxmox::Compute::Disks`  and storage.volumes are volumes so it returns `Fog::Proxmox::Compute::Volumes`. You can remove the volumes without caring about server but you cannot remove the disk without powering off the server (hotplug issue).

stejskalleos (approved):
It's been up for a while and has been used and tested well,
so let's get this in finally.

Manisha15:
Thank you so much!! @stejskalleos

PR #92, opened 2020-04-07
Detach ISO after build

sbernhard:
Detach the ISO Image after build. Attach the ISO image and mark with "start_connected" is necessary so that rebuild the host works (otherwise vmware would simply disconnect the iso if you reboot the host)

### DIFF
--- a/app/models/concerns/foreman_bootdisk/orchestration/compute.rb
+++ b/app/models/concerns/foreman_bootdisk/orchestration/compute.rb
@@ -9,7 +9,6 @@ module ForemanBootdisk
 
       included do
         after_validation :queue_bootdisk_compute
-        after_build :rebuild_with_bootdisk
       end
 
       def bootdisk_isodir
@@ -21,15 +20,25 @@ module ForemanBootdisk
       end
 
       def queue_bootdisk_compute
-        return unless compute? && errors.empty? && new_record?
+        return unless compute? && errors.empty?
         return unless provision_method == 'bootdisk'
 
-        queue.create(name: _('Generating ISO image for %s') % self, priority: 5,
-                     action: [self, :setGenerateIsoImage])
-        queue.create(name: _('Upload ISO image to datastore for %s') % self, priority: 6,
-                     action: [self, :setIsoImage])
-        queue.create(name: _('Attach ISO image to CDROM drive for %s') % self, priority: 1001,
-                     action: [self, :setAttachIsoImage])
+        # We want to add our queue jobs only if really necessary:
+        #   - in case of starting to create a new host
+        #   - in case of a rebuild
+        if (old.nil? || !old.build?) && build?
+          queue.create(name: _('Generating ISO image for %s') % self, priority: 5,
+                       action: [self, :setGenerateIsoImage])
+          queue.create(name: _('Upload ISO image to datastore for %s') % self, priority: 6,
+                       action: [self, :setIsoImage])
+          queue.create(name: _('Attach ISO image to CDROM drive for %s') % self, priority: 1001,
+                       action: [self, :setAttachIsoImage])
+        # Detach ISO image when host sends 'built' HTTP POST request
+        elsif old&.build? && !build?
+          queue.create(name: _('Detach ISO image from CDROM drive for %s') % self, priority: 52,
+                       action: [self, :setDetachIsoImage])
+        end
+        true
       end
 
       def bootdisk_generate_iso_image
@@ -44,6 +53,10 @@ module ForemanBootdisk
 
       def bootdisk_attach_iso
         compute_resource.iso_attach(File.basename(bootdisk_isofile), uuid)
+      end
+
+      def bootdisk_detach_iso
+        compute_resource.iso_detach(uuid)
       end
 
       def setGenerateIsoImage
@@ -73,18 +86,14 @@ module ForemanBootdisk
 
       def delAttachIsoImage; end
 
-      def rebuild_with_bootdisk
-        return true unless bootdisk?
-
-        begin
-          bootdisk_generate_iso_image
-          bootdisk_upload_iso
-          bootdisk_attach_iso
-        rescue StandardError => e
-          Foreman::Logging.exception "Failed to rebuild Bootdisk image for #{name}", e, level: :error
-          return false
-        end
+      def setDetachIsoImage
+        logger.info format('Detaching ISO image from CDROM drive for %s', name)
+        bootdisk_detach_iso
+      rescue StandardError => e
+        failure format(_('Failed to detach ISO image from CDROM drive of instance %{name}: %{message}'), name: name, message: e.message), e
       end
+
+      def delDetachIsoImage; end
     end
   end
 end
--- a/test/unit/concerns/orchestration/compute_test.rb
+++ b/test/unit/concerns/orchestration/compute_test.rb
@@ -22,29 +22,48 @@ module ForemanBootdisk
       @host.send(:setAttachIsoImage)
     end
 
-    test 'provisioning a host with provision method bootdisk should queue bootdisk tasks' do
+    test 'provisioning a host with provision method bootdisk should detach iso' do
+      @cr.expects(:iso_detach)
+      @host.send(:setDetachIsoImage)
+    end
+
+    test 'provisioning a new host with provision method bootdisk should queue bootdisk tasks' do
       @host.stubs(:compute?).returns(true)
+      @host.stubs(:build?).returns(true)
       @host.send(:queue_bootdisk_compute)
       tasks = @host.queue.all.map(&:name)
       assert_includes tasks, "Generating ISO image for #{@host.name}"
       assert_includes tasks, "Upload ISO image to datastore for #{@host.name}"
       assert_includes tasks, "Attach ISO image to CDROM drive for #{@host.name}"
+      assert_not_includes tasks, "Detach ISO image from CDROM drive for #{@host.name}"
     end
 
-    test 'should rebuild bootdisk' do
-      @host.expects(:bootdisk_generate_iso_image).returns(true)
-      @host.expects(:bootdisk_upload_iso).returns(true)
-      @host.expects(:bootdisk_attach_iso).returns(true)
-      assert @host.rebuild_with_bootdisk
+    test 'rebuilding a host with provision method bootdisk should queue bootdisk tasks' do
+      @host.stubs(:compute?).returns(true)
+      old = stub()
+      old.stubs(:build?).returns(false)
+      @host.stubs(:old).returns(old)
+      @host.stubs(:build?).returns(true)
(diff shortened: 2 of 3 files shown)


app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -46,6 +47,16 @@ def bootdisk_attach_iso
         compute_resource.iso_attach(File.basename(bootdisk_isofile), uuid)
       end
 
+      def bootdisk_detach_iso
+        return true unless provision_method == 'bootdisk'
+        return true if previous_changes['build'].nil?
+
+        if previous_changes['build'] == [true, false]
```
timogoebel, line 54: I don't quite get this, shouldn't `bootdisk_detach_iso` only run `before_provision` anyway?

Why don't you make this part of the orchestration queue?
E.g. modifying `queue_bootdisk_compute` to queue a detach task when the host leaves build mode? There should be [plenty of examples](https://github.com/theforeman/foreman/blob/c86ed9c68adedaaff5562ddaa6e7229ec416427d/app/models/concerns/orchestration/dhcp.rb#L156) that demonstrate the required code changes.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -85,6 +96,14 @@ def rebuild_with_bootdisk
           return false
         end
       end
+
+      private
+      def is_rebuild?
```
timogoebel, line 101: Nitpick: This is concern is included into Host::Managed, e.g. this will provide `host.is_rebuild?`.  Maybe something like `host.left_build_mode?` or similar?

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -71,10 +82,10 @@ def setAttachIsoImage
         failure format(_('Failed to attach ISO image to CDROM drive of instance %{name}: %{message}'), name: name, message: e.message), e
       end
 
-      def delAttachIsoImage; end
```
timogoebel, line 74: I believe this needs to stay there or a rollback action won't work in case an error occurs.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
sbernhard: I had the same idea. I added this to foreman_bootdisk to finish this task sooner. :-)

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
sbernhard: regarding orchestration queue: detaching should be done before starting the provisioned system. I guess, its not possible to add a task right before that stage. "before provision" is executed after the host has signalized the "built" flag (in the build snippet). Therefore the procedure should be:
- in provisioning template, the host "eject" the cdrom
- in provisioning template the snippet build runs and sends the built flag
- bootdisk_detach_iso is used to finally remove the ISO

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
sbernhard: done

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
timogoebel: When the host requests the built callback, the queue also runs. I believe this is used e.g. for tftp orchestration to switch the host from pxe boot to local boot.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
timogoebel: it can stay in the plugin, I just think the naming is misleading.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
sbernhard: I guess you mean this one: https://github.com/theforeman/foreman/blob/bad5f878b6fb275dbc7267930bcd0cb589d9aa66/app/models/concerns/orchestration.rb#L13 This is actually run to the wrong point in time.. :(

timogoebel: No, I mean the default queue.

sbernhard: I can not find out how to specify when a task should be run when using the queue. Of course, there are priorities but this isn't really "safe".

lzap: As Timo said, the best place to learn would be TFTP orchestration when Foreman flips TFTP config back to boot from local drive. They are hooked via `after_validation :queue_tftp` and then it compares the `built` flag. If it was updated, it schedules redeployment of files. Isn't this exactly what you are looking for?

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -72,19 +91,6 @@ def setAttachIsoImage
       end
 
       def delAttachIsoImage; end
```
timogoebel, line 93: This needs to call `bootdisk_detach_iso` for a proper rollback, doesn't it?

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -46,6 +55,16 @@ def bootdisk_attach_iso
         compute_resource.iso_attach(File.basename(bootdisk_isofile), uuid)
       end
 
+      def bootdisk_detach_iso
+        return true unless provision_method == 'bootdisk'
+        return true if previous_changes['build'].nil?
+
+        if previous_changes['build'] == [true, false]
```
timogoebel, line 62: I believe we can remove these guards in this method. A developer should make sure to only call these methods when the host is in the correct state.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -46,6 +55,16 @@ def bootdisk_attach_iso
...
+          logger.info format('Detaching ISO image from CDROM drive for %s', name)
+          compute_resource.iso_detach(uuid)
+        end
+      end
```
timogoebel, line 66:
```suggestion
      def bootdisk_detach_iso
        logger.info format('Detaching ISO image from CDROM drive for %s', name)
        compute_resource.iso_detach(uuid)
      end
```

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -21,15 +20,25 @@ def bootdisk_isofile
...
+        #   - in case of a rebuild
+        if (old.nil? || !old.build?) && build?
+          queue.create(name: _('Generating ISO image for %s') % self, priority: 5,
+                       action: [self, :setGenerateIsoImage])
+          queue.create(name: _('Upload ISO image to datastore for %s') % self, priority: 6,
+                       action: [self, :setIsoImage])
+          queue.create(name: _('Attach ISO image to CDROM drive for %s') % self, priority: 1001,
+                       action: [self, :setAttachIsoImage])
+        end
+
```
timogoebel, line 37: I think a comment is in order here as well.
```suggestion

#  if build mode is switched off
```

sbernhard:
Please, have a look again.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -21,15 +20,25 @@ def bootdisk_isofile
...
+        #   - in case of starting to create a new host
+        #   - in case of a rebuild
+        if (old.nil? || !old.build?) && build?
+          queue.create(name: _('Generating ISO image for %s') % self, priority: 5,
+                       action: [self, :setGenerateIsoImage])
+          queue.create(name: _('Upload ISO image to datastore for %s') % self, priority: 6,
+                       action: [self, :setIsoImage])
+          queue.create(name: _('Attach ISO image to CDROM drive for %s') % self, priority: 1001,
+                       action: [self, :setAttachIsoImage])
+        # Detach ISO image when host sends 'built' HTTP POST request
```
timogoebel, line 36: Actually this would also run when the user clicks "Cancel build" in the UI.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
```diff
@@ -21,15 +20,25 @@ def bootdisk_isofile
...
+        elsif old&.build? && !build?
+          queue.create(name: _('Detach ISO image from CDROM drive for %s') % self, priority: 52,
```
timogoebel, line 38: What happens if the iso image has been detached manually by the user, e.g. via the vCenter UI?

test/unit/concerns/orchestration/compute_test.rb
```diff
@@ -22,29 +22,48 @@ class OrchestrationComputeTest < ActiveSupport::TestCase
       @host.send(:setAttachIsoImage)
     end
 
-    test 'provisioning a host with provision method bootdisk should queue bootdisk tasks' do
+    test 'provisioning a host with provision method bootdisk should detach iso' do
```
timogoebel, line 25: Is this test any good?

test/unit/concerns/orchestration/compute_test.rb
```diff
@@ -22,29 +22,48 @@ class OrchestrationComputeTest < ActiveSupport::TestCase
...
     end
 
-    test 'should skip rebuild bootdisk' do
-      host = FactoryBot.build(:host,
-                              compute_resource: @cr)
-      host.expects(:bootdisk_generate_iso_image).never
-      host.expects(:bootdisk_upload_iso).never
-      host.expects(:bootdisk_attach_iso).never
-      assert host.rebuild_with_bootdisk
+    test 'detaching the bootdisk should be done' do
```
timogoebel, line 55: Nit:

```suggestion
    test 'the iso should be detached when the host leaves build mode' do
```

test/unit/concerns/orchestration/compute_test.rb
sbernhard: added this one because there were 2 tests (iso_upload, iso_attach) above testing the methods directly.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
sbernhard: Just tested this. 
Detaching a "disconnected" device results in a successful request.

app/models/concerns/foreman_bootdisk/orchestration/compute.rb
sbernhard: ... which would be more a feature than a bug as it would detach the ISO image again :)

PR #93, opened 2020-04-07
Fixes #4413 - UEFI support for full and generic

lzap:
This builds on top of the following PRs:

* https://github.com/theforeman/foreman_bootdisk/pull/33
* https://github.com/theforeman/foreman_bootdisk/pull/62
* https://github.com/theforeman/foreman_bootdisk/pull/64

It takes the best of all. For full host bootdisk, simply EFI floppy image is generated with Grub2 template. It reuses what whas built for iPXE (kernel/initrd paths). This is straightforward and works just fine (it worked from the day one).

For generic image, the idea is to use Grub2 in EFI to directly download Grub2 template from Foreman via HTTP. Controllers and rake tasks will need to pass Grub2 template to generators.

DO NOT MERGE YET the generic bootdisk is not yet finished, I am investigating why grub2 goes to black screen the moment it tries to chainload.


app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +88,35 @@ def self.generate(opts = {})
...
           ipxe_source_file = File.join(Setting[:bootdisk_ipxe_dir], 'ipxe.lkrn')
-          raise(Foreman::Exception, N_('Please ensure the ipxe-bootimgs package is installed.')) unless File.exist?(ipxe_source_file)
+          raise Foreman::Exception.new(N_('Please ensure the ipxe-bootimgs package is installed.')) unless File.exist?(ipxe_source_file)
 
           FileUtils.cp(ipxe_source_file, File.join(wd, 'build', 'ipxe'))
           File.open(File.join(wd, 'build', 'script'), 'w') { |file| file.write(opts[:ipxe]) }
         end
 
+        if opts[:grub]
+          grub_path = "/var/lib/tftpboot/grub2/grubx64.efi"
```
timogoebel, line 98: This is quite a scary method, maybe this can make it a little better.

```ruby
FileUtils.mkdir_p(File.join(wd, 'build', 'EFI', 'BOOT'))

grub_files = ['grubx64.efi', 'shimx64.efi']

copy_grub2_files(files: grub_files, to: wd)
```

and then

```ruby
def copy_grub2_files(files:, to:)
  files.each do |file|
    from = File.join('/var/lib/tftpboot/grub2/', file)
    raise Foreman::Exception.new(N_('Ensure %{path} exists by enabling TFTP smart proxy module.'), path: from) unless File.exist?(from)
    FileUtils.cp(from, File.join(to, 'build', 'EFI', 'BOOT', file))
  end
end

Not complete, but I guess you get the idea...

Comment (lzap) [reply app/services/foreman_bootdisk/iso_generator.rb:98]: Well, sometimes simple copy is more readable than complex looping that can be error prone. But I can do it, I don't like extra method tho. By the way `File.join('/var/lib/tftpboot/grub2/', file)` is actually only half-way there, but let's do that. I hate File.join use - Foreman will never run natively on Windows. If ever, WSL uses proper UNIX paths... :-)

File: app/services/foreman_bootdisk/iso_generator.rb [hunk 2]
```diff
@@ -76,12 +88,35 @@ def self.generate(opts = {})
...
+          shim_path = "/var/lib/tftpboot/grub2/shimx64.efi"
+          raise Foreman::Exception.new(N_('Ensure %{path} exists by enabling TFTP smart proxy module.'), path: grub_path) unless File.exist?(grub_path)
```
adamruzicka, line 100: This exception was raised even though I had the tftp smart proxy module enabled. Either the message or puppet module/packaging needs to be fixed. Installing `foreman-bootloaders-redhat-tftpboot` made it go away.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +88,35 @@ def self.generate(opts = {})
...
+          File.open(File.join(wd, 'build', 'EFI', 'BOOT', 'grub.cfg'), 'w') { |file| file.write(opts[:grub]) }
+          efibootimg = File.join(wd, 'build', 'efiboot.img')
+          system("mformat -T 7000 -C -i #{efibootimg}")
+          system("mmd -i #{efibootimg} '::/EFI'")
+          system("mmd -i #{efibootimg} '::/EFI/BOOT'")
+          result_grub = system("mcopy -m -i #{efibootimg} '#{grub_path}' '::/EFI/BOOT/grubx64.efi'")
+          result_shim = system("mcopy -m -i #{efibootimg} '#{shim_path}' '::/EFI/BOOT/BOOTX64.efi'")
+          mkiso_args = "-eltorito-alt-boot -e efiboot.img -no-emul-boot"
+          isohybrod_args = ["--uefi"]
+          raise Foreman::Exception.new(N_('Bootdisk EFI full, increase number of tracks')) unless (result_grub && result_shim)
```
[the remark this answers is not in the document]
timogoebel: Can we calculate the required amount of tracks?

app/services/foreman_bootdisk/iso_generator.rb
lzap: Okay, after another refactoring I have to say that few copies of lines are sometimes better and this is the case. I introduced a bug when I was modifying the code and had to debug it simply because it was much harder to read. I am putting it back it was.

timogoebel: Ok, works for me. Can you cover the bug with a test case instead? ;-)

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +88,35 @@ def self.generate(opts = {})
...
+          grub_path = "/var/lib/tftpboot/grub2/grubx64.efi"
+          shim_path = "/var/lib/tftpboot/grub2/shimx64.efi"
+          raise Foreman::Exception.new(N_('Ensure %{path} exists by enabling TFTP smart proxy module.'), path: grub_path) unless File.exist?(grub_path)
+          raise Foreman::Exception.new(N_('Ensure %{path} exists by enabling TFTP smart proxy module.'), path: shim_path) unless File.exist?(shim_path)
+          FileUtils.mkdir_p(File.join(wd, 'build', 'EFI', 'BOOT'))
+          FileUtils.cp(grub_path, File.join(wd, 'build', 'EFI', 'BOOT', 'grubx64.efi'))
+          FileUtils.cp(shim_path, File.join(wd, 'build', 'EFI', 'BOOT', 'shimx64.efi'))
+          File.open(File.join(wd, 'build', 'EFI', 'BOOT', 'grub.cfg'), 'w') { |file| file.write(opts[:grub]) }
+          efibootimg = File.join(wd, 'build', 'efiboot.img')
+          system("mformat -T 7000 -C -i #{efibootimg}")
```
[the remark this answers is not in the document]
lzap: I created a helper method and all system calls have error checking now. I tested `truncate` method and it works fine in libvirt, since ESP partitions are simple MSDOS FAT partitions this should work fine.

app/services/foreman_bootdisk/iso_generator.rb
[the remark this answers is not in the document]
lzap: Removed creation of the image using the floppy tool, looks like we can simply create an empty file and format it.

app/services/foreman_bootdisk/iso_generator.rb
lzap: This RPM is only for downstream, running installer should do it: https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/tftp/netboot.pp

lzap: Anyway, I created new setting so this path is configurable. It will try to find grub in `/boot/efi/EFI` first and TFTP directory will be a fallback. This should now work out of box even when you don't have TFTP module enabled.

lzap:
Rebased, incorporated your concerns. The patch is now ready for review and test.

* Full host bootdisk works fine, fully supported.
* Generic host disks are implemented too, however during my testing I experienced corrupted downloads of kernel or init ram disk. This is filed as https://bugzilla.redhat.com/show_bug.cgi?id=1822600 and after this is resolved it should work too. There is no way to test this end-to-end, so please test that the disk generates fine. I added a 30 seconds delay with big warning that this will not work at the moment.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -24,42 +25,60 @@ def self.generate_full_host(host, opts = {}, &block)
...
         err = host.errors.full_messages.to_sentence
-        raise(::Foreman::Exception.new(N_('Unable to generate disk PXELinux template: %s'), err))
+        raise Foreman::Exception.new(N_('Unable to generate disk %{kind} template: %{error}'), kind: kind, error: err)
       end
 
       template
     end
 
+    def self.system_or_exception(command, error)
+      unless system(command)
```
timogoebel, line 51: In another PR we should refactor this to properly capture and log (for debugging purposes) the output of these commands.

```ruby
require 'open3'
stdout, stderr, status = Open3.capture3(command)
logger.debug "Whatever #{stdout}, ..."
```

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +95,34 @@ def self.generate(opts = {})
...
+          grub_cfg = File.join(wd, 'build', 'EFI', 'BOOT', 'grub.cfg')
+          File.open(grub_cfg, 'w') { |file| file.write(opts[:grub]) }
+          efibootimg = File.join(wd, 'build', 'efiboot.img')
+          system_or_exception("truncate -s 4M #{efibootimg}", N_('Creating new image failed, install truncate utility'))
+          system_or_exception("mkfs.msdos #{efibootimg}", N_('Failed to format the ESP image via mkfs.msdos'))
+          system_or_exception("mmd -i #{efibootimg} '::/EFI'", N_('Failed to create a directory within the ESP image'))
+          system_or_exception("mmd -i #{efibootimg} '::/EFI/BOOT'", N_('Failed to create a directory within the ESP image'))
+          [['grubx64.efi', 'GRUBX64.EFI'], ['shimx64.efi', 'BOOTX64.EFI']].each do |pair|
+            src = File.join(Setting[:bootdisk_grub2_dir], pair.first)
+            dst = "::/EFI/BOOT/#{pair.second}"
```
timogoebel, line 115:
```suggestion
          {'grubx64.efi' => 'GRUBX64.EFI', 'shimx64.efi' => 'BOOTX64.EFI'}.each do |source_file, destination_file|
            src = File.join(Setting[:bootdisk_grub2_dir], source_file)
            dst = "::/EFI/BOOT/#{destination_file}"
```

app/services/foreman_bootdisk/iso_generator.rb
lzap: Agreed, perhaps for all commands issued. Maybe something in Foreman core even.

app/services/foreman_bootdisk/iso_generator.rb
lzap: Done.

lzap:
Fixed tests, rebased, this is now ready for testing.

lzap:
Found the problem, it is the BOOTIF what breaks the full host disk. It cannot be set to `BOOTIF=01-$mac_default_net` because when not PXE/HTTP booting, this variable is NOT set in grub, therefore it renders to `BOOTIF=01-` which is wrong and dracut cannot initialize the network. Going to fix that.

lzap:
https://github.com/theforeman/community-templates/pull/738

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -24,42 +25,60 @@ def self.generate_full_host(host, opts = {}, &block)
...
-        raise(::Foreman::Exception.new(N_('Unable to generate disk PXELinux template: %s'), err))
+        raise Foreman::Exception.new(N_('Unable to generate disk %{kind} template: %{error}'), kind: kind, error: err)
       end
 
       template
     end
 
+    def self.system_or_exception(command, error)
+      unless system(command)
+        logger.debug "Bootdisk command failed: #{command}"
```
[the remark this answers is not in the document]
lzap: Thanks for reminder, fixing.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +95,34 @@ def self.generate(opts = {})
...
+          efibootimg = File.join(wd, 'build', 'efiboot.img')
+          system_or_exception("truncate -s 4M #{efibootimg}", N_('Creating new image failed, install truncate utility'))
+          system_or_exception("mkfs.msdos #{efibootimg}", N_('Failed to format the ESP image via mkfs.msdos'))
+          system_or_exception("mmd -i #{efibootimg} '::/EFI'", N_('Failed to create a directory within the ESP image'))
+          system_or_exception("mmd -i #{efibootimg} '::/EFI/BOOT'", N_('Failed to create a directory within the ESP image'))
+          {'grubx64.efi' => 'GRUBX64.EFI', 'shimx64.efi' => 'BOOTX64.EFI'}.each do |source_file, destination_file|
+            src = File.join(Setting[:bootdisk_grub2_dir], source_file)
+            dst = "::/EFI/BOOT/#{destination_file}"
+            raise(Foreman::Exception.new(N_('Ensure %{path} contains grubx64.efi and shimx64.efi or change path in settings'), path: src)) unless File.exist?(src)
+            raise(Foreman::Exception.new(N_('Unable to mcopy %{path}'), path: src)) unless system("mcopy -m -i #{efibootimg} '#{src}' '#{dst}'")
```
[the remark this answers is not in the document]
lzap: Thanks for reminder, fixing.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +95,34 @@ def self.generate(opts = {})
...
+          File.open(grub_cfg, 'w') { |file| file.write(opts[:grub]) }
+          efibootimg = File.join(wd, 'build', 'efiboot.img')
+          system_or_exception("truncate -s 4M #{efibootimg}", N_('Creating new image failed, install truncate utility'))
+          system_or_exception("mkfs.msdos #{efibootimg}", N_('Failed to format the ESP image via mkfs.msdos'))
+          system_or_exception("mmd -i #{efibootimg} '::/EFI'", N_('Failed to create a directory within the ESP image'))
+          system_or_exception("mmd -i #{efibootimg} '::/EFI/BOOT'", N_('Failed to create a directory within the ESP image'))
+          {'grubx64.efi' => 'GRUBX64.EFI', 'shimx64.efi' => 'BOOTX64.EFI'}.each do |source_file, destination_file|
+            src = File.join(Setting[:bootdisk_grub2_dir], source_file)
+            dst = "::/EFI/BOOT/#{destination_file}"
+            raise(Foreman::Exception.new(N_('Ensure %{path} contains grubx64.efi and shimx64.efi or change path in settings'), path: src)) unless File.exist?(src)
```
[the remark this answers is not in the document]
lzap: Damn, this one makes it a little bit more complex. What I am going to do is to use `/var/lib/tftpboot/grub2` and if it does not exists it would fall back to `/var/lib/foreman/bootdisk` path and recommend the user to do this in the exception:

Ensure %{path} contains grubx64.efi and shimx64.efi: install TFTP module or copy those files into /var/lib/foreman/bootdisk (Grub2 directory setting)

lzap:
Ok pushing the changes you found:

```diff
diff --git a/app/models/setting/bootdisk.rb b/app/models/setting/bootdisk.rb
index 58d9b3d..1502e10 100644
--- a/app/models/setting/bootdisk.rb
+++ b/app/models/setting/bootdisk.rb
@@ -6,7 +6,7 @@ class Setting
       ipxe = ['/usr/lib/ipxe'].find { |p| File.exist?(p) } || '/usr/share/ipxe'
       isolinux = ['/usr/lib/ISOLINUX'].find { |p| File.exist?(p) } || '/usr/share/syslinux'
       syslinux = ['/usr/lib/syslinux/modules/bios', '/usr/lib/syslinux'].find { |p| File.exist?(p) } || '/usr/share/syslinux'
-      grub2 = ['/boot/efi/EFI/fedora', '/boot/efi/EFI/redhat', '/boot/efi/EFI/centos', '/boot/efi/EFI/debian', '/boot/efi/EFI/ubuntu'].find { |p| File.exist?(p) } || '/var/lib/tftpboot/grub2'
+      grub2 = ['/var/lib/tftpboot/grub2'].find { |p| File.exist?(p) } || '/var/lib/foreman/bootdisk'
       templates = -> { Hash[ProvisioningTemplate.where(template_kind: TemplateKind.where(name: 'Bootdisk')).map { |temp| [temp[:name], temp[:name]] }] }
 
       [
diff --git a/app/services/foreman_bootdisk/iso_generator.rb b/app/services/foreman_bootdisk/iso_generator.rb
index cb3fe48..b90d797 100644
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -49,8 +49,8 @@ module ForemanBootdisk
 
     def self.system_or_exception(command, error)
       unless system(command)
-        logger.debug "Bootdisk command failed: #{command}"
-        Foreman::Exception.new(N_(error))
+        ForemanBootdisk.logger.debug "Bootdisk command failed: #{command}"
+        raise Foreman::Exception.new(N_(error))
       end
     end
 
@@ -103,8 +103,10 @@ module ForemanBootdisk
 
         if opts[:grub]
           FileUtils.mkdir_p(File.join(wd, 'build', 'EFI', 'BOOT'))
-          grub_cfg = File.join(wd, 'build', 'EFI', 'BOOT', 'grub.cfg')
-          File.open(grub_cfg, 'w') { |file| file.write(opts[:grub]) }
+          # a copy when using CDROM directly
+          File.open(File.join(wd, 'build', 'grub.cfg'), 'w') { |file| file.write(opts[:grub]) }
+          # a copy when using USB
+          File.open(File.join(wd, 'build', 'EFI', 'BOOT', 'grub.cfg'), 'w') { |file| file.write(opts[:grub]) }
           efibootimg = File.join(wd, 'build', 'efiboot.img')
           system_or_exception("truncate -s 4M #{efibootimg}", N_('Creating new image failed, install truncate utility'))
           system_or_exception("mkfs.msdos #{efibootimg}", N_('Failed to format the ESP image via mkfs.msdos'))
@@ -113,7 +115,7 @@ module ForemanBootdisk
           {'grubx64.efi' => 'GRUBX64.EFI', 'shimx64.efi' => 'BOOTX64.EFI'}.each do |source_file, destination_file|
             src = File.join(Setting[:bootdisk_grub2_dir], source_file)
             dst = "::/EFI/BOOT/#{destination_file}"
-            raise(Foreman::Exception.new(N_('Ensure %{path} contains grubx64.efi and shimx64.efi or change path in settings'), path: src)) unless File.exist?(src)
+            raise(Foreman::Exception.new(N_('Ensure %{path} contains grubx64.efi and shimx64.efi: install TFTP module or copy those files into /var/lib/foreman/bootdisk (Grub2 directory setting)'), path: src)) unless File.exist?(src)
             raise(Foreman::Exception.new(N_('Unable to mcopy %{path}'), path: src)) unless system("mcopy -m -i #{efibootimg} '#{src}' '#{dst}'")
           end
           mkiso_args = "-eltorito-alt-boot -e efiboot.img -no-emul-boot"
```

Now I am going to test iPXE bootdisks with the correct grub2 version to see how that looks like.

lzap:
Ok it's done, tests will fail and I will work on them tomorrow but feel free to start final review and testing.

For EFI this works:

* Full Host Bootdisk
* Subnet Generic Bootdisk

The other two bootdisk generates an error and halt the system when booted over EFI. It's because HTTP EFI BOOT can only be used against proxy (therefore subnet is required). Also this makes it easier to support as both bootdisk are two extremes: full host disk is good for preregistered hosts and generic works for all of them.

Full host bootdisk works out of box and nothing is required.

Subnet bootisk requires:

 * SUBNET GENERIC IMAGE ONLY (host image not supported)
 * HOST PARAM default_grub_install_entry set to efi_http
 * PROXY WITH HTTPBOOT FEATURE
 * HTTP UEFI BOOT ONLY (Legacy/PXE not supported)
 * IPv4 ONLY (IPv6 not tested, change the template)
 * HTTP ONLY (change the template for HTTPS)
 * ISC DHCP (other servers not tested)
 * GRUB FROM RHEL 8.3+/7.9+ (when generating the image)
 * EFI HTTP or HTTPS grub entry must be selected in menu

Also make sure to modify bootdisk templates via seeding or manually (the *.erb templates in this PR).

I recommend to test with grub2 from upciming RHEL 8.3 or 7.9, I can send the link for download as I experienced a regression in Fedora Rawhide. Should be fixed now but to stay on the safe side.

The following PRs are required in order to get this working and we need them in 2.1 too:

* https://github.com/theforeman/community-templates/pull/738
* https://github.com/theforeman/foreman/pull/7713

lzap:
One more comment, I got reply from Javier from the grub team and I am setting prefix incorrectly here, it should be:

```
set root=(http,192.168.122.1:80)
set http_path=/
```

Then the extra grub entry for HTTP/HTTPS would not be even necessary. I need to test this tho.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +99,40 @@ def self.generate(opts = {})
...
-          raise(Foreman::Exception, N_('Please ensure the ipxe-bootimgs package is installed.')) unless File.exist?(ipxe_source_file)
+          raise Foreman::Exception.new(N_('Please ensure the ipxe-bootimgs package is installed.')) unless File.exist?(ipxe_source_file)
 
           FileUtils.cp(ipxe_source_file, File.join(wd, 'build', 'ipxe'))
           File.open(File.join(wd, 'build', 'script'), 'w') { |file| file.write(opts[:ipxe]) }
         end
 
+        if opts[:grub]
+          # a copy when using CDROM directly
+          File.open(File.join(wd, 'build', 'grub.cfg'), 'w') { |file| file.write(opts[:grub]) }
```
adamruzicka, line 110: When booting in UEFI, grub looks for `/efi/boot/grub.cfg` with `/` being the device it booted from. When I attached the bootdisk image as a cdrom, I got into grub prompt.
```suggestion
          FileUtils.mkdir_p(File.join(wd, 'build', 'EFI', 'BOOT'))
          File.open(File.join(wd, 'build', 'EFI', 'BOOT', 'grub.cfg'), 'w') { |file| file.write(opts[:grub]) }
```

lzap: Yeah I have found this as well, fixed.

lzap:
I pushed the final version, tested all 4 scenarios and it works fine. Make sure to have both patches and seed the latest template, there were many changes in it. The only known workaround is that PXEGrub2 template must be in the absolute form of `(http,ip_address:port)/absolute/path/to/kernel` to solve two issues (both reported for grub team):

* inability to send DHCP request with DNS option (therefore DNS does not work)
* inability to carry over `http_path` variable which is needed to be able to find the files on the correct path

Thus the extra option we need to carry until this is fixed.

Tests TBD, please review now I am in time pressure @adamruzicka @timogoebel @ekohl and keep in mind testing this after bigger changes is probably an hour of work so be nice, the whole plugin is a hack so only important issues please :-)

app/controllers/foreman_bootdisk/api/v2/subnet_disks_controller.rb
```diff
@@ -24,9 +24,10 @@ def index; end
         param :subnet_id, :identifier_dottable, required: true
         def subnet
           subnet = @subnet_disk
-          subnet.tftp || raise(::Foreman::Exception.new(N_('TFTP feature not enabled for subnet %s'), subnet.name))
-          tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
-          ForemanBootdisk::ISOGenerator.generate(ipxe: tmpl) do |iso|
+          subnet.tftp || subnet.httpboot || raise(::Foreman::Exception.new(N_('TFTP or HTTPBOOT feature not enabled for subnet %s'), subnet.name))
```
lzap, line 27: I should probably check for `subnet.template` as well as it's mandatory now.

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -65,7 +65,7 @@ def bootdisk_button_disabled(host)
 
     # need to wrap this one in a test for template proxy presence
     def display_bootdisk_for_subnet(host)
-      if (proxy = host.try(:subnet).try(:tftp)) && proxy.has_feature?('Templates')
+      if (proxy = host.try(:subnet).try(:tftp) || host.try(:subnet).try(:httpboot)) && proxy.has_feature?('Templates')
```
lzap, line 68: Here it is actually being required in order to show the buttons, so the test should definitely be there.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -24,42 +25,64 @@ def self.generate_full_host(host, opts = {}, &block)
...
+      if opts[:grub].nil? && opts[:ipxe]
+        opts[:grub] = <<~EOT
+          echo "Grub2 template not associated and/or unsupported bootdisk type."
+          echo "The following bootdisks are supported for EFI systems:"
+          echo " * FULL HOST BOOTDISK"
+          echo " * SUBNET GENERIC BOOTDISK"
+          echo "The system will poweroff in few minutes..."
+          sleep 500
+          poweroff
+        EOT
```
lzap, line 77: This message appears when you generate on of the other bootdisk types (it will not raise an error) but boot it on EFI instead of BIOS system.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -76,12 +99,42 @@ def self.generate(opts = {})
...
+          }.each do |src, dest|
+            raise(Foreman::Exception.new(N_('Ensure %{path} contains grubx64.efi and shimx64.efi: install TFTP module or copy those files into /var/lib/foreman/bootdisk (Grub2 directory setting)'), path: src)) unless File.exist?(src)
+            raise(Foreman::Exception.new(N_('Unable to mcopy %{path}'), path: src)) unless system("mcopy -m -i #{efibootimg} '#{src}' '#{"::/EFI/BOOT/#{dest}"}'")
+          end
+          mkiso_args = "-eltorito-alt-boot -e efiboot.img -no-emul-boot"
+          isohybrod_args = ["--uefi"]
+        else
+          mkiso_args = ''
+          isohybrod_args = []
+        end
```
lzap, line 136: This damn block took me two weeks of finding out how things work, emails with grub developers and countless failed boots... :beer:

app/views/foreman_bootdisk/generic_efi_host.erb
```diff
@@ -0,0 +1,59 @@
...
+<%
+# possible values are: "http" or "https"
+proxy_proto = "http"
+
+@subnet || bootdisk_raise("Generic disk not supported for EFI, use subnet disk")
+@subnet.httpboot? || bootdisk_raise("Requires a proxy with httpboot feature")
+@subnet.template? || bootdisk_raise("Requires a proxy with template feature")
+proxy_port = @subnet.httpboot.setting(:HTTPBoot, "#{proxy_proto}_port") || bootdisk_raise("Requires proxy with #{proxy_proto}_port exposed setting")
+proxy_httpboot_host = @subnet.httpboot.hostname
+proxy_template_host = @subnet.template.hostname
```
adamruzicka, line 30: Don't we need to resolve the hostnames to ipaddresses the same way as we do here https://github.com/theforeman/community-templates/blob/f2f58d2c67ce8d47fde09f78f91142682403c224/provisioning_templates/PXEGrub2/kickstart_default_pxegrub2.erb#L39 ?

lzap: We absolutely do, my smart proxy had IP address since I was testing this on devel setup.

lzap:
Oh I forgot, please merge core and templates PR so we can have that in RC3 and I am going to fix tests now.

app/controllers/foreman_bootdisk/api/v2/subnet_disks_controller.rb
```diff
@@ -25,10 +25,15 @@ def index; end
...
-          tmpl = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
-          ForemanBootdisk::ISOGenerator.generate(ipxe: tmpl) do |iso|
-            name = subnet.name
-            send_data File.read(iso), filename: "bootdisk_subnet_#{name}.iso"
+          tmpl_bios = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
+          tmpl_efi = nil
+          if subnet.httpboot
+            tmpl_efi = ForemanBootdisk::Renderer.new.generic_efi_template_render(subnet)
+          else
+            ForemanBootdisk.logger.warn('HTTPBOOT feature not enabled for subnet %s, UEFI not be available for bootdisk' % subnet.name)
```
timogoebel, line 33:
```suggestion
            ForemanBootdisk.logger.warn('HTTPBOOT feature is not enabled for subnet %s, UEFI may not be available for bootdisk' % subnet.name)
```

app/lib/foreman_bootdisk/scope/bootdisk.rb
```diff
@@ -18,6 +17,10 @@ def bootdisk_chain_url(mac = host.try(:mac), action = 'iPXE')
       def bootdisk_raise(*args)
         raise ::Foreman::Exception.new(*args)
       end
+
+      def template_name
```
timogoebel, line 21: I believe this is a method that should be defined for the source, not the actual template scope.

@kamils-ironin: Do you remember?

app/services/foreman_bootdisk/renderer.rb
```diff
@@ -4,36 +4,46 @@
...
 
+    def generic_template_render(subnet = nil)
+      render_template(template: generic_host_template, host: stub_host(subnet), subnet: subnet)
+    end
+
+    def generic_efi_template_render(subnet)
+      if subnet.httpboot?
+        render_template(template: generic_efi_host_template, host: stub_host(subnet), subnet: subnet)
+      else
+        ForemanBootdisk.logger.warn('HTTPBOOT feature not enabled for subnet %s, UEFI not be available for bootdisk' % subnet.name)
```
timogoebel, line 27:
```suggestion
        ForemanBootdisk.logger.warn('HTTPBOOT feature is not enabled for subnet %s, UEFI may not be available for bootdisk' % subnet.name)
```

app/lib/foreman_bootdisk/scope/bootdisk.rb
kamils-iRonin: I think you are right, the template name is taken from the source

https://github.com/theforeman/foreman/blob/7b0e8889cbff82e6b5caf1e09752f17d0dbe3ffc/lib/foreman/renderer/scope/base.rb#L26

app/controllers/foreman_bootdisk/api/v2/subnet_disks_controller.rb
lzap: Fixed them all (4 in total).

app/services/foreman_bootdisk/renderer.rb
lzap: Fixed them all (4 in total).

app/lib/foreman_bootdisk/scope/bootdisk.rb
lzap: I need to do it this way because bootdisk actually also renders PXE Linux templates using the default scopes.

lzap:
Rebased, tests are not passing because some database issues, hopefully they will run now.

Also added this:

```diff
diff --git a/app/services/foreman_bootdisk/iso_generator.rb b/app/services/foreman_bootdisk/iso_generator.rb
index 48d0e72..c495503 100644
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -92,7 +92,7 @@ module ForemanBootdisk
 
           FileUtils.cp(isolinux_source_file, File.join(wd, 'build', 'isolinux.bin'))
 
-          source_files = ['ldlinux.c32', 'menu.c32']
+          source_files = ['ldlinux.c32', 'menu.c32', 'libutil.c32']
           source_files.each do |source_file|
             full_path = File.join(Setting[:bootdisk_syslinux_dir], source_file)
             FileUtils.cp(full_path, File.join(wd, 'build', source_file)) if File.exist?(full_path)
```

To workaround a fullhost bootdisk warning "Failed to load libutil.c32". It worked but the file should probably be there.

lzap:
https://community.theforeman.org/t/plugin-db-drop-tests-fail-with-the-production-database-is-not-configured-for-the-production-environment/19161

lzap:
There was one problem, I've refactored all controllers to use `send_file` instead `send_data` to save memory. However this action is asynchronous and offloaded to web server which is a problem for temporary directory created using `mktmpdir` with block. I've changed the method not to use block so the temp directory is not cleaned up after send file is initiated so it works fine. Drawback: temp files are no longer cleaned up, but thanks to systemd this will be cleaned automatically: https://www.freedesktop.org/software/systemd/man/systemd-tmpfiles-setup.service.html

lzap:
All bootdisks load fine, tested UI, CLI and rake tasks. Once tests are green I am merging.

adamruzicka:
Thank you @lzap !

lzap:
https://github.com/theforeman/foreman-packaging/pull/5400

lzap:
For the record, I've sent one commend which was supposed to land here into a different PR:

@adamruzicka I updated the tests, but there is one snag. One controller test is failing with "ActionController::MissingExactTemplate: ForemanBootdisk::DisksController#subnet is missing a template for request formats: text/html" and I don't know why. I use the same stubbing as for other two:

    @controller.expects(:send_file).with('/tmp/fake.iso', anything).returns(:success)

Yet it works for the others and not for the one. If I understand the problem correctly, Rails expect controllers to call "render" method however since `send_file` is stubbed it does not get called therefore Rails tries to figure the content out.

One way to solve this would be to actually create a temp file and let the `send_file` do the job but that's clunky. Maybe you can spot why this works for the other two controllers.

https://github.com/theforeman/smart-proxy/pull/758

PR #95, opened 2020-07-13
Fixes #30390 - Add DSL docs for bootdisk plugin

ofedoren:


ares:
oh, I'd merge if I had permissions, @lzap would you mind?

PR #96, opened 2020-08-13
Fixes #30759 - fix http-proxy issue

m-bucher:
Fixes the issue described [here](https://community.theforeman.org/t/global-http-proxy-does-not-work-in-bootdisk/14273).

I'd like to have a generic solution in foreman as everyone else, but it does not really look as that will come soon.

### DIFF
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -10,6 +10,8 @@ require 'uri'
 # requires syslinux, ipxe/ipxe-bootimgs, genisoimage, isohybrid
 module ForemanBootdisk
   class ISOGenerator
+    extend Foreman::HTTPProxy
+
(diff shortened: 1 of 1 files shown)


lzap (approved):

lzap:
Thank you, can you create a new RM issue and associate this PR with it. You can link it with my original RM issue, but that one I would like to be a tracker for some final solution.

Also, there is a test failing:

```
ForemanBootdisk::IsoGeneratorTest::#generate_full_host#test_0001_fetch handles redirect:
NameError: uninitialized constant Foreman::HTTPproxy
```

m-bucher:
@lzap I am pretty sure that adding the necessary dependency for testing to the normal dependencies is not correct.
However, the jenkins run explicitly does a `bundle install --without development`.
Is there another way to add the `minitest-stub-const`-gem for testing only?

lzap (approved):
Thank you.

PR #98, opened 2020-09-16
Add Rubocop Github Action job

lzap:
This is an attempt to move Rubocop from CI to GHAction

### DIFF
--- a/.github/workflows/rubocop.yaml
+++ b/.github/workflows/rubocop.yaml
@@ -1,0 +1,19 @@
+name: Rubocop
+
+on: [pull_request]
+
+jobs:
+  rubocop:
+    runs-on: ubuntu-latest
+    env:
+      BUNDLE_FOREMAN: '0'
+      BUNDLE_WITHOUT: development:test
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.5
+          bundler-cache: true
+      - name: Run rubocop
+        run: bundle exec rubocop -P --format github
--- a/Gemfile
+++ b/Gemfile
@@ -2,17 +2,6 @@
 
 source 'http://rubygems.org'
 
+gem 'theforeman-rubocop', '~> 0.1.0.pre', groups: %i[development rubocop]
+
 gemspec
-
-FOREMAN_GEMFILE = File.expand_path('.foreman_app/Gemfile', __dir__)
-if File.exist?(FOREMAN_GEMFILE)
-  instance_eval(Bundler.read_file(FOREMAN_GEMFILE))
-else
-  puts <<~MESSAGE
-    Foreman source code is not present. To get the latest version, run:
-
-      rake setup:foreman_prepare
-
-    and try again.
-  MESSAGE
-end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,23 +1,22 @@
-require:
-  - rubocop-performance
-  - rubocop-rails
-  - rubocop-minitest
+inherit_gem:
+  theforeman-rubocop:
+    - lenient.yml
 
 AllCops:
-  TargetRubyVersion: 2.5
-  TargetRailsVersion: 5.2
+  NewCops: disable
   Exclude:
     - 'release-gem'
     - 'test/**/*'
     - 'lib/tasks/**/*'
-
-Rails:
-  Enabled: true
+    - 'node_modules/**/*'
+    - 'tmp/**/*'
+    - 'templates/**/*'
+    - 'vendor/**/*'
 
 Style:
   Enabled: false
 
-Layout/LineLength:
+Layout:
   Enabled: false
 
 Metrics:
--- a/foreman_bootdisk.gemspec
+++ b/foreman_bootdisk.gemspec
@@ -24,9 +24,8 @@ Gem::Specification.new do |s|
   ]
   s.files = `git ls-files`.split("\n") - Dir['.*', 'Gem*', '*.gemspec']
 
-  s.add_development_dependency 'rubocop', '~> 0.80.0'
-  s.add_development_dependency 'rubocop-minitest', '~> 0.7.0'
-  s.add_development_dependency 'rubocop-performance', '~> 1.5.2'
-  s.add_development_dependency 'rubocop-rails', '~> 2.4.2'
+  s.required_ruby_version = '>= 2.5'
+
+  s.add_development_dependency 'theforeman-rubocop'
   s.add_development_dependency 'webmock'
 end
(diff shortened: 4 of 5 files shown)


.github/workflows/rubocop.yaml
```diff
@@ -0,0 +1,19 @@
...
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: RuboCop Linter Action
+        uses: andrewmcodes/rubocop-linter-action@v3.3.0
```
ekohl, line 17: https://github.com/andrewmcodes/rubocop-linter-action#rubocop-linter-action states:
> PLEASE DO NOT USE THIS ACTION, and instead follow [the Rails way](https://github.com/rails/rails/blob/4a78dcb326e57b5035c4df322e31875bfa74ae23/.github/workflows/rubocop.yml#L1).

lzap: Interesting, adopted Rails job. It's unfortunate not to see those nice comments but this will deliver a good experience too. It's super fast at least.

Gemfile
```diff
@@ -3,16 +3,3 @@
...
-  instance_eval(Bundler.read_file(FOREMAN_GEMFILE))
-else
-  puts <<~MESSAGE
-    Foreman source code is not present. To get the latest version, run:
-
-      rake setup:foreman_prepare
-
-    and try again.
-  MESSAGE
-end
```
lzap, line 18: This is no longer in our template, honestly I don't know what is this and if this ever worked. There is no Rakefile in this repo so the instructions are not correct.

https://github.com/theforeman/foreman_plugin_template/blob/master/Gemfile

.github/workflows/rubocop.yaml
ekohl: The comments don't work on PRs from other repos, which is in practice all PRs.

lzap:
@ekohl is this acceptable? Can I proceed with deleting the rake task call on our CI? I wonder if I can do this per plugin repo.

ekohl:
We can run Rubocop in Jenkins as well and solve it for every plugin that uses Jenkins, but since there was basically no plugin maintainer response on my RFC I've concluded that plugin maintainers don't want any central support so I'm fine with everyone taking care of their own things.

lzap:
Is there any change required in foreman-infra? I am trying to undrstand how all of this works, I think the script named `puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh` is the correct one to do however I don't see anything related to rubocop there. And Foreman core`Rakefile` appears to be empty, weird.

ekohl:
That's because plugins now typically add their Rubocop to the `jenkins:unit` rake task. So for this plugin it's here:
https://github.com/theforeman/foreman_bootdisk/blob/8ab8807715e2377f2c969a4b8f1740edc52827a4/lib/tasks/bootdisk.rake#L110

lzap:
Ok I have removed this from bootdisk then, my understanding is that this is causing issues to Proxmox.

ekohl:
> my understanding is that this is causing issues to Proxmox.

No, the issue in https://github.com/theforeman/foreman_bootdisk/pull/90 is that **foreman_fog_proxmox** needs to be modified. Removing it here will only help plugins that depend on foreman_bootdisk.

.github/workflows/rubocop.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby 2.7
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.7
+    - name: Cache gems
+      uses: actions/cache@v1
```
ezr-ondrej, line 16:
```suggestion
      uses: actions/cache@v2
```
V1 won't work for us (with the pattern `restore-keys`)

lzap:
Let's hold this PR until @ezr-ondrej is done with core Rubocop changes and ability to share rules and then we can move from there.

.github/workflows/rubocop.yaml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby 2.7
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.7
```
ezr-ondrej, line 14: Confident much? xD
I feel it's pretty new ruby and foreman is not ready at all :thinking:

.github/workflows/rubocop.yaml
```diff
@@ -0,0 +1,27 @@
+name: Rubocop
+
+on: [push, pull_request]
```
ezr-ondrej, line 3:
```suggestion
on: [pull_request]
```

ezr-ondrej (approved):
Few nits, loving it in general :+1:

.github/workflows/rubocop.yaml
lzap: I am running my dev setup on 2.7 for quite some time now. Granted I see a lot of warnings but 2.7.2 dropped these I think.

lzap:
Ok rebased. I implemented `all.yaml` from the shared gem with some tuning :-)

sbernhard:
> Let's hold this PR until @ezr-ondrej is done with core Rubocop changes and ability to share rules and then we can move from there.

Is there a link to the planned rubocop change?

lzap:
This is ready for re-review, let's finally merge it and unblock ATIX guys with that proxmox PR?

lzap:
> Is there a link to the planned rubocop change?

It is still WIP but I think we are good to proceed here. I am going to rebase and fix remaining problems.

https://community.theforeman.org/t/rubocop-common-styles-in-gem/20794

lzap:
Success, this is now using pre-release of the foreman-rubocop which also supports github commenting.

lzap:
@sbernhard would you be find with being nominated for bootdisk push access? You are the most active engineer in the last year and @timogoebel is busy so another member with push access would be great.

ezr-ondrej (approved):
Looks great @lzap !

lzap:
@timogoebel would you mind?

timogoebel (approved):

timogoebel:
Thanks!

sbernhard:
> @sbernhard would you be find with being nominated for bootdisk push access? You are the most active engineer in the last year and @timogoebel is busy so another member with push access would be great.

sure! I would do my best :)

lzap:
https://community.theforeman.org/t/nominating-bernhard-suttner-for-push-access-to-foreman-bootdisk/23144

PR #100, opened 2021-01-07
Fixes: #31585 iPXE ping gateway and name server

jcpunk:
Attempt to ping critical network resources.  In theory when everything is working this should be transparent to the users.  If not, this will clearly identify what went wrong.

### DIFF
--- a/app/views/foreman_bootdisk/generic_efi_host.erb
+++ b/app/views/foreman_bootdisk/generic_efi_host.erb
@@ -44,6 +44,10 @@ net_bootp
 net_ls_addr
 net_ls_routes
 net_ls_dns
+echo Trying to ping proxy_httpboot_host: <%= proxy_httpboot_host %>
+ping --count 1 <%= proxy_httpboot_host %> || echo Ping to proxy_httpboot_host failed or ping command not available.
+echo Trying to ping proxy_template_host: <%= proxy_template_host %>
+ping --count 1 <%= proxy_template_host %> || echo Ping to proxy_template_host failed or ping command not available.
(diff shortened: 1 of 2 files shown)


lzap:
Hello,

thanks for the patch.

First of all, when iPXE command fails it immediately exits execution thus this will lead to boot failures on systems where `ping` command is not avilable. To workaround this, use `||` or shell-like operator.

Second, both gateway and DNS server are variables available within iPXE environment as long as network was initialized.

Then there is also a problem with your syntax, according to docs it's `--timeout N` not `--timeout=N`, I haven`t tested if your syntax is supported or not. I'd rather stick with what's documented.

Please do this instead (`show` command is actually not needed but I find it useful):

```
show ${netX/gateway}
ping --count 1 ${netX/gateway} || echo Ping to gateway failed!
show ${netX/dns}
ping --count 1 ${netX/dns} || echo Ping to dns failed!
```

Can you also add these four lines into Kickstart default iPXE, Preseed default iPXE, AutoYaST default iPXE and Windows default iPXE right before `kernel/initrd` commands?

FYI variable `netX` is set to the last initialized network interface, this should work in all contexts.

jcpunk:
I'll get this updated with your requested changes.

> Can you also add these four lines into Kickstart default iPXE, Preseed default iPXE, AutoYaST default iPXE and Windows default iPXE right before kernel/initrd commands?

I'm showing those templates in `foreman` itself, I'll open a PR over there with those additions.

lzap (approved):
Thanks, all is green I can smash the button for this one.

PR #101, opened 2021-03-01
Refs #30759 - Fix decission on http-proxy usage

m-bucher:
Looks like I got the method wrong in the first fix.


lzap (approved):
Yeah this is not ideal, but thanks for the fix.

Sorry for the delay, I will be cutting a new release for 2.4 with this patch this week.

PR #102, opened 2021-03-09
Fixes #32057 - delete old temporary bootdisks

lzap:
This is because we now use PrivateTmp and systemd only cleans temp files after the service is stopped. But bootdisks are big (50MB each) and users will often fill their disks. The solution is to use permanent temp (/var/tmp) and clean them via systemd-tmpfiles.

A followup PRs into packaging will be associated with this PR, do not merge until packaging changes pass reviews.

### DIFF
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -85,8 +85,13 @@ module ForemanBootdisk
         EOT
       end
 
-      # Temporary directory cannot be cleaned inprocess due to send_file offloaded to web server
-      wd = Dir.mktmpdir('bootdisk-iso-')
+      # Temporary directory cannot be cleaned in-process due to send_file offloaded to web server and
+      # the call is asynchronous. Also $TMPDIR (/tmp or /var/tmp) cannot be used because Foreman runs
+      # with PrivateTmp systemd setting. And the plugin does not ship its own systemd-tmpfiles config,
+      # so let's delete old temporary files explicitly for files older than 30 minutes:
+      Dir.glob(Rails.root.join('tmp/clean-daily/bootdisk-iso-*')).select{|f| File.ctime(f) < (Time.now - (60*30)) }.each{|f| FileUtils.rm_f(f)}
+      # And create new temporary directory:
+      wd = Dir.mktmpdir('bootdisk-iso-', Rails.root.join('tmp/clean-daily'))
       Dir.mkdir(File.join(wd, 'build'))
 
       if opts[:isolinux]
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -85,13 +85,8 @@ module ForemanBootdisk
         EOT
       end
 
-      # Temporary directory cannot be cleaned in-process due to send_file offloaded to web server and
-      # the call is asynchronous. Also $TMPDIR (/tmp or /var/tmp) cannot be used because Foreman runs
-      # with PrivateTmp systemd setting. And the plugin does not ship its own systemd-tmpfiles config,
-      # so let's delete old temporary files explicitly for files older than 30 minutes:
-      Dir.glob(Rails.root.join('tmp/clean-daily/bootdisk-iso-*')).select{|f| File.ctime(f) < (Time.now - (60*30)) }.each{|f| FileUtils.rm_f(f)}
       # And create new temporary directory:
-      wd = Dir.mktmpdir('bootdisk-iso-', Rails.root.join('tmp/clean-daily'))
+      wd = Dir.mktmpdir('bootdisk-iso-', Rails.root.join('tmp'))
       Dir.mkdir(File.join(wd, 'build'))
 
       if opts[:isolinux]
@@ -168,6 +163,14 @@ module ForemanBootdisk
       raise Foreman::Exception.new(N_('ISO hybrid conversion failed: %s'), $?) unless system(*["isohybrid", isohybrod_args, iso].flatten.compact)
 
       yield iso
+    ensure
+      # Clean the working directory (not the ISO file itself)
+      FileUtils.rm_f(File.join(wd, 'build'))
+      # Temporary directory cannot be cleaned in-process due to asynchronous send_file call.
+      # Also we cannot rely on systemd-tmpfiles-clean as private temporary files are not subject
+      # of scheduled cleanups. Let's clean bootdisks from prevous requests manually by finding
+      # and deleting all directories created 30 minutes ago.
+      Rails.root.glob('tmp/bootdisk-iso-*').select { |f| File.ctime(f) < (Time.now.to_i - (60 * 30)) }.each { |f| FileUtils.rm_f(f) }
     end
 
     def self.build_mkiso_command(output_file:, source_directory:, extra_commands:)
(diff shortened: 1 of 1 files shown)


lzap:
After some discussion, I will rewrite this patch to cleanup old temp directories manually in the codebase. I want a simple solution for now.

lzap:
This is now ready for review, @timogoebel

timogoebel (approved):
Looks good enough!

@ekohl: Any thoughts, perhaps?

lzap:
For the record, Ewoud suggested that instead of send_file we could use send_stream and I agree however we have a BZ for this one and I will be likely backporting this one as we treat "disk out of space" bugs as important to fix in Satellite. So I am seeking for a small fix in this case.

timogoebel:
@lzap: I believe this is good enough for now. Can you take a look at the CI, then we can merge this.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -85,8 +85,13 @@ def self.generate(opts = {})
         EOT
       end
 
-      # Temporary directory cannot be cleaned inprocess due to send_file offloaded to web server
-      wd = Dir.mktmpdir('bootdisk-iso-')
+      # Temporary directory cannot be cleaned in-process due to send_file offloaded to web server and
+      # the call is asynchronous. Also $TMPDIR (/tmp or /var/tmp) cannot be used because Foreman runs
+      # with PrivateTmp systemd setting. And the plugin does not ship its own systemd-tmpfiles config,
+      # so let's delete old temporary files explicitly for files older than 30 minutes:
+      Dir.glob(Rails.root.join('tmp/bootdisk-iso-*')).select{|f| File.ctime(f) < (Time.now - (60*30)) }.each{|f| FileUtils.rm_f(f)}
```
ekohl, line 92: I think this also works since Rails.root is a [Pathname](https://ruby-doc.org/stdlib-2.6.3/libdoc/pathname/rdoc/Pathname.html):
```suggestion
      Rails.root.glob('tmp/bootdisk-iso-*').select{|f| File.ctime(f) < (Time.now - (60*30)) }.each{|f| FileUtils.rm_f(f)}
```

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -85,8 +85,13 @@ def self.generate(opts = {})
         EOT
       end
 
-      # Temporary directory cannot be cleaned inprocess due to send_file offloaded to web server
-      wd = Dir.mktmpdir('bootdisk-iso-')
+      # Temporary directory cannot be cleaned in-process due to send_file offloaded to web server and
+      # the call is asynchronous. Also $TMPDIR (/tmp or /var/tmp) cannot be used because Foreman runs
+      # with PrivateTmp systemd setting. And the plugin does not ship its own systemd-tmpfiles config,
```
ekohl, line 90: I don't understand this reasoning. Why can't `$TMPDIR` be used because it's on `PrivateTmp`? To me it doesn't appear related at all. On Debian you have pretty much the same guarantees for `Rails.root.join('tmp')` as for `$TMPDIR`: it only exists while the service is running and is destroyed on service restart. On Red Hat we do place a tmpfiles entry so it is persisted between runs.

Also, offloading `send_file` to the web server probably doesn't work since both `$TMPDIR` and `Rails.root.join('tmp')` live in directories that only the Foreman user can read so I think sendfile support doesn't work. I also don't think sendfile works with a reverse proxy setup.

So really, I think the commend should be "this allows for an easy cherry pick"

ekohl (commented):
Not very familiar with the code, but it looks like `generate` creates a bunch of temporary files and then combines it into a single ISO file. That iso file is yielded. Can't `generate` at ensure all those temporary files are cleaned up? Then you at least clean up most files immediately and saves you probably 50%. It also doesn't require a next request to clean up the files.

app/services/foreman_bootdisk/iso_generator.rb
lzap: > Why can't $TMPDIR be used because it's on PrivateTmp?

That is just a bad wording. This is what I mean:

```
        # Temporary directory cannot be cleaned in-process due to asynchronous send_file call.
        # Also we cannot rely on systemd-tmpfiles-clean as private temporary files are not subject
        # of scheduled cleanups. Let's clean bootdisks from prevous requests manually.
```

> Also, offloading send_file to the web server probably doesn't work

But it works. Frankly, I gave it about 15 minutes or research and I don't understand how this is possible - there is not a single mention of send_file in Puma, Rails appears to call `send_file` from `response` (Rack object?) and in Rack documentation there are few words about selected http servers: https://www.rubydoc.info/gems/rack/Rack/Sendfile and it appears they need to be configured in a special way which we do not do I suppose.

But it works, there must be some kind of fallback mechanism. I just tried it on a 2.4 RC3 production install to be sure.

> So really, I think the commend should be "this allows for an easy cherry pick"

Well, this is an unfair comment. I have volunteered to maintain the repository but I never had ambitions to refactor it in a big way. I am trying to keep this plugin alive.

There are two options: `send_file` which is known to work in production for years, and `send_data` which would allow synchronous call (thus cleanup after the download) but it would require the Rails process to load the whole file into memory so it is probably harder on memory. There is also an option to stream the data manually via another Rails API but since I am already having issues in the other (core) PR with that, I am really hesitant to start anything like this here.

lzap:
> Can't generate at ensure all those temporary files are cleaned up?

Good idea I can probably do that.

lzap:
I rebased.

app/services/foreman_bootdisk/iso_generator.rb
ekohl: > But it works, there must be some kind of fallback mechanism. I just tried it on a 2.4 RC3 production install to be sure.

I'd think it falls back to sending the file the old fashioned way: send bytes from a Ruby process.

lzap:
Amended to fix the rubocop issue.

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -163,6 +163,14 @@ def self.generate(opts = {})
...
 
       yield iso
+    ensure
+      # Clean the working directory (not the ISO file itself)
+      FileUtils.rm_f(File.join(wd, 'build'))
+      # Temporary directory cannot be cleaned in-process due to asynchronous send_file call.
+      # Also we cannot rely on systemd-tmpfiles-clean as private temporary files are not subject
+      # of scheduled cleanups. Let's clean bootdisks from prevous requests manually by finding
+      # and deleting all directories created 30 minutes ago.
+      Rails.root.glob('tmp/bootdisk-iso-*').select { |f| File.ctime(f) < (Time.now.to_i - (60 * 30)) }.each { |f| FileUtils.rm_f(f) }
```
ekohl, line 173: Maybe it should store the time in a variable rather than recalculating it inside the loop?
```suggestion
      delete_older_than = Time.now.to_i - (60 * 30)
      Rails.root.glob('tmp/bootdisk-iso-*').select { |f| File.ctime(f) < delete_older_than }.each { |f| FileUtils.rm_f(f) }
```

ekohl (approved):

lzap:
Sure thank, would you mind squashing on merge? Thanks.

timogoebel:
Thanks!

PR #104, opened 2021-03-24
fixes #31726- Introducing disabling types of bootdisk

domitea:
This PR introduces ability to disable types of bootdisk in settings

### DIFF
--- a/app/controllers/concerns/allowed_actions.rb
+++ b/app/controllers/concerns/allowed_actions.rb
@@ -1,0 +1,16 @@
+module AllowedActions
+  extend ActiveSupport::Concern
+
+  def bootdisk_type_allowed?(action = params[:action])
+    return true if Setting::Bootdisk.allowed_types&.include?(action)
+
+    message = _('This type of bootdisk is not allowed. Please contact administrator.')
+    if api_request?
+      render_error :custom_error, status: :unprocessable_entity, locals: { message: message}
+    else
+      error(message)
+      redirect_back(fallback_location: '/')
+    end
+    false
+  end
+end
--- a/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
@@ -7,11 +7,13 @@ module ForemanBootdisk
     module V2
       class DisksController < ::Api::V2::BaseController
         include ::Api::Version2
+        include AllowedActions
 
         resource_description do
           api_base_url '/bootdisk/api'
         end
 
+        before_action :bootdisk_type_allowed?, only: :generic
         before_action :find_resource, only: :host
         skip_after_action :log_response_body
 
@@ -34,10 +36,12 @@ module ForemanBootdisk
         def host
           host = @disk
           if params[:full]
+            return unless bootdisk_type_allowed?('full_host')
             ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
               send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
             end
           else
+            return unless bootdisk_type_allowed?
             # EFI not supported for iPXE host bootdisk
             tmpl = host.bootdisk_template_render
             ForemanBootdisk::ISOGenerator.generate(ipxe: tmpl) do |iso|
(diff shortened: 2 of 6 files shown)


app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -48,6 +51,12 @@ def host
 
         private
 
+        def action_is_allowed(action = params[:action])
+          return true if Setting.find_by_name('bootdisk_allowed_types').value.include? action
```
domitea, line 55: There is the exactly same code in the `subnet_disks_controller.rb` so it leads to extract this method to a concern probably. However, this only one method so I don't know if it's worth it.

ares (changes requested):
Looks reasonable but I think the UI should also not display the buttons if the bootdisk is disabled. User shouldn't click on the button to only find out it's disabled. The protection on the controller level should of course remain too.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
lzap: Isn't writing a helper method in settings/bootdisk.rb more simple?

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -34,10 +35,12 @@ def generic
         def host
           host = @disk
           if params[:full]
+            return unless action_is_allowed('full_host')
```
lzap, line 38: This should be symbol.

app/models/setting/bootdisk.rb
```diff
@@ -21,12 +21,17 @@ def self.default_settings
...
       ]
     end
 
     def self.humanized_category
       N_('Boot disk')
     end
+
+    def self.bootdisk_types
+      ['generic', 'host', 'full_host', 'subnet']
+    end
```
lzap, line 35: It's a good practice to use symbols in Ruby, memory management of symbols is way more efficient than strings and they are also immutable.

I suggest to create a helper method here which would do all the logic: `bootdisk_type_allowed?(type_as_symbol)`. This way you can simplify most of the code in this patch.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
ares: yeah, that would probably also work and logically belongs there, :+1:

app/controllers/concerns/allowed_actions.rb
```diff
@@ -0,0 +1,16 @@
+module AllowedActions
+  extend ActiveSupport::Concern
+
+  def bootdisk_type_allowed?(action = params[:action])
+    return true if Setting.find_by(name: 'bootdisk_allowed_types').value.include?(action.to_sym)
```
lzap, line 5: Any particular reason not to use Setting["XYZ"] here? Also you can create a helper method in the Setting class named with the same name as here and call it.

lzap (changes requested):
The menu is broken: 

image

app/controllers/concerns/allowed_actions.rb
lzap: This resolves to strings here:

```
[1] pry(main)> Setting.find_by(name: 'bootdisk_allowed_types').value
=> ["generic", "host", "full_host", "subnet"]
```

I think our settings might not even support symbols to store data. My previous comment still applies too.

lzap (changes requested):
Does not work, I see no bootdisks.

app/models/setting/bootdisk.rb
```diff
@@ -21,12 +21,18 @@ def self.default_settings
...
+        set('bootdisk_allowed_types', N_('List of allowed bootdisk types, remove type to disable it'), Setting::Bootdisk.bootdisk_types, N_('Allowed bootdisk types'))
       ]
     end
 
     def self.humanized_category
       N_('Boot disk')
     end
+
+    def self.bootdisk_types
+      [:generic, :host, :full_host, :subnet]
```
lzap, line 34: I know I wanted you to have symbols here, but in this case we actually want strings because this is confusing - in the end, settings stack will convert these to strings anyway so let's not trick reader here.

lzap:
Failures are now relevant.

lzap (approved):
It works fine, thanks.

PR #105, opened 2021-04-01
fixes #10814, #11254 - Moving link to subnet page and introducing subnet disk controller

domitea:
This PR solves two big problems in bootdisk plugin. Both of them are tied with subnet. The first one moves button for download subnet image on subnet page. That's a huge improvement because right know this button lays at the detail of host page. This has historical reason becuase subnet image generation has been at disks controller for host. This is a second problem, so the logic of generation of subnet image is moves to separate controller.

### DIFF
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -5,9 +5,11 @@ require 'uri'
 module ForemanBootdisk
   class DisksController < ::ApplicationController
     include AllowedActions
+    include PrettyError
 
     before_action :bootdisk_type_allowed?, except: :help
-    before_action :find_resource, only: %w[host full_host subnet]
+
+    before_action :find_resource, only: %w[host full_host]
 
     # as this engine is isolated, we need to include url helpers from core explicitly
     # to render help page layout
@@ -51,38 +53,12 @@ module ForemanBootdisk
       end
     end
 
-    def subnet
-      host = @disk
-      begin
-        subnet = host.try(:subnet) || raise(::Foreman::Exception.new(N_('Subnet is not assigned to the host %s'), host.name))
-        subnet.tftp || raise(::Foreman::Exception.new(N_('TFTP feature not enabled for subnet %s'), subnet.name))
-        subnet.httpboot || ForemanBootdisk.logger.warn('HTTPBOOT feature is not enabled for subnet %s, UEFI may not be available for bootdisk' % subnet.name)
-
-        tmpl_bios = ForemanBootdisk::Renderer.new.generic_template_render(subnet)
-        tmpl_efi = ForemanBootdisk::Renderer.new.generic_efi_template_render(subnet)
-      rescue StandardError => e
-        error_rendering(e)
-        redirect_back(fallback_location: '/')
-        return
-      end
-
-      ForemanBootdisk::ISOGenerator.generate(ipxe: tmpl_bios, grub: tmpl_efi) do |iso|
-        send_file(iso, filename: "bootdisk_subnet_#{subnet.name}.iso")
-      end
-    end
-
     def help; end
 
     private
 
     def resource_scope(_controller = controller_name)
       Host::Managed.authorized(:view_hosts)
-    end
-
-    def error_rendering(exception)
-      msg = _('Failed to render boot disk template')
-      error("#{msg}: #{exception.message}")
-      ::Foreman::Logging.exception(msg, exception)
     end
   end
 end
--- a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
+++ b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
@@ -2,16 +2,11 @@
 
 module ForemanBootdisk
   module HostsHelperExt
+    include BootdiskLinksHelper
+
     def host_title_actions(host)
       if host.bootdisk_downloadable?
-        title_actions(
-          button_group(
-            select_action_button(
-              _('Boot disk'), { class: 'btn btn-group' },
-              host_action_buttons(host)
-            )
-          )
-        )
+        bootdisk_title_action_buttons(host_action_buttons(host))
       else
         bootdisk_button_disabled(host)
       end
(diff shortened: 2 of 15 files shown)


domitea:
I added another PR to core, that is necessary for correct behavior because it adds a pagelet endpoint for button at subnet index page.

(See https://github.com/theforeman/foreman/pull/8425)

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -4,6 +4,8 @@
 
 module ForemanBootdisk
   class DisksController < ::ApplicationController
+    include PrettyError
+
     before_action :find_resource, only: %w[host full_host subnet]
```
ares, line 9: since subnet is removed, the before action list should be also updated

lzap:
Thanks, as I've said please ping me if you don't see my activity in 24 hours.

The code looks good to me, please fix Rubocop and I can start testing.

lzap:
Cool tests are passing testing now.

lzap:
We need to move Generic image as well. Generic image is similar to Subnet image, except it does connect directly to Foreman Server instead of Proxy. Host screen is not the best place, however Subnet page also feels a bit, maybe just a little bit more appropriate. I think we should move it to the Subnet page just like we moved Subnet image. What you think @ares ?

If you click on Host - Bootdisk - Help there is the following screen (see the picture below). We can easily customize the page - if we add two titles: Host images (the first two) and Generic images (subnet + generic) we can also add a sentence to each chapter where users can download them. If we add the same Help link to the Subnet page, I think it will be pretty easy for users to find both Subnet and Generic images.

Snímek obrazovky 2021-04-09 v 14 25 50

lzap:
Hmm I do not see anything for my subnet, what is wrong?

image

lib/foreman_bootdisk/engine.rb
```diff
@@ -57,6 +58,10 @@ class Engine < ::Rails::Engine
         provision_method 'bootdisk', N_('Boot disk based')
         template_labels 'Bootdisk' => N_('Boot disk embedded template')
         allowed_template_helpers :bootdisk_chain_url, :bootdisk_raise
+
+        extend_page "subnets/index" do |cx|
+          cx.add_pagelet :action_buttons, name: "Bootdisk", partial: "subnets/bootdisk_button"
```
lzap, line 63: I think `action_buttons` is not a good symbol to name a pagelet. There are many "action buttons" in the app, this should have been something like `subnet_index_action_buttons`. Would you mind opening another core PR (just do refs issue number) fixing this while we can?

lzap (changes requested):
Existing host bootdisks do work fine, however the buttons do not show up on the subnet page.

domitea:
I've updated PR with you commetns @lzap. The endpoint for pagelet is renamed and I also add new endpoint for help and generic image - https://github.com/theforeman/foreman/pull/8441. 
I think that I understand your effort to move generic image to subnet also. So there is a new button at subnet index page at title buttons. There should be bootdisk button (like the host one) with link to generic image and help. I think that's better place to generic image, because it's a host agnostic and subnet agnostic image, I can't to connect them with any subnet at all. 

What do you think about that?

domitea:
`HostJSTest` is failing due to connection refused errror. So I think that it's not relevant for this PR.

lzap:
I just merged the core patch make sure to update your branch.

You updated the help screen but I do not see the Generic image moved into a Subnet page, could you do it? At least I still see it on a Host page.

Also add a Help button to the Subnet page too.

ares:
I don't mind moving the generic disk as well, bonus points for updating the plugin documentation with the information where to find it.

ares:
Since it now depends on the core PR that has been added in 2.5, the plugin engine should specify the minimum Foreman version and this should also mean bumping the plugin major version.

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -69,55 +64,30 @@ def host_action_buttons(host)
...
       full_host_image_link = display_bootdisk_link_if_authorized(
                                _("Full host '%s' image") % host.name.split('.')[0],
-                               {
-                                   controller: 'foreman_bootdisk/disks',
-                                   action: 'full_host',
-                                   id: host
+                             {
+                                 controller: 'foreman_bootdisk/disks',
+                                 action: 'full_host',
+                                 id: host
```
lzap, line 80: Just an advice, try to stay away from these reformatting commits. On big projects like Foreman, everyone has a preference and this usually ends up in never ending stream of hunks fighting each other. Unless you want to introduce a Rubocop rule to enforce something, in that case do it in a separate PR.

Just for the future, you can keep it not big deal.

app/views/foreman_bootdisk/disks/help.html.erb
```diff
@@ -9,28 +9,38 @@
     <%= _('All images are usable as either ISOs or as disk images, including being written to a USB disk with `dd`.') %>
   </p>
 
-  <h2><%= _('Host image') %></h2>
+  <h2><%= _('Host images') %></h2>
+  <p>
+    <%= _('This images are used for host. You can find them at host detail page.') %>
```
lzap, line 14: These images...

The same down below.

app/views/subnets/_bootdisk_title_buttons.erb
```diff
@@ -0,0 +1 @@
+<%= display_bootdisk_title_buttons %>
```
lzap, line 1: I wonder if the code for both templates could actually live here and you would get rid of an extra helper. But I am not a RoR expert, not sure if that's not a good practice. Sounds reasonable to me tho.

lzap:
What Marek said about the engine.rb, please update. Then I can re-test. Thanks, great work so far!

app/views/subnets/_bootdisk_title_buttons.erb
domitea: I learned that every unique piece of code should be in project only once (Don't repeat yourself - DRY). So since these helper has very similar codebase, it makes sense to move same functionality to another place and you just include them in desired helpers. 

So I can get rid of extra helper but there will be almost same code twice and I don't think it's a good idea. What do you think?

domitea:
I've updated PR with your comments. Please check that the updated versioning are correct.

app/views/subnets/_bootdisk_title_buttons.erb
lzap: Okay DRY is good sure.

lzap:
Rubocop failure.

lzap:
If you could just rebase to get it passing, this looks like some failure at github.com.

lzap (approved):
It works great, thank you.

PR #106, opened 2021-04-09
fixes #31674 - Updating token expiration when bootdisk is downloaded

domitea:
This PR tries to solves problem where Bootdisk produces useless bootdisk
by hardcoding tokens after it's expiration. So before the bootdisk is
downloaded, the token for host is updated with new expiration time.

### DIFF
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -41,6 +41,8 @@ module ForemanBootdisk
         return
       end
 
+      prolong_token(host)
+
       ForemanBootdisk::ISOGenerator.generate(ipxe: tmpl) do |iso|
         send_file(iso, filename: "#{host.name}.iso")
       end
@@ -48,6 +50,9 @@ module ForemanBootdisk
 
     def full_host
       host = @disk
+
+      prolong_token(host)
+
       ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
         send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
       end
@@ -59,6 +64,20 @@ module ForemanBootdisk
 
     def resource_scope(_controller = controller_name)
       Host::Managed.authorized(:view_hosts)
+    end
+
+    def error_rendering(exception)
+      msg = _('Failed to render boot disk template')
+      error("#{msg}: #{exception.message}")
+      ::Foreman::Logging.exception(msg, exception)
+    end
+
+    def prolong_token(host)
+      return if Setting[:token_duration] == 0 || host.token.nil?
+
+      # update build token
+      token = host.token
+      token.update(expires: Time.zone.now + Setting[:token_duration].minutes)
     end
   end
 end
--- a/test/functional/foreman_bootdisk/disks_controller_test.rb
+++ b/test/functional/foreman_bootdisk/disks_controller_test.rb
@@ -69,6 +69,27 @@ class ForemanBootdisk::DisksControllerTest < ActionController::TestCase
     end
   end
 
+  describe '#host without tftp' do
+    setup :setup_referer
+    setup :setup_org_loc
+    setup :setup_subnet_no_tftp
+    setup :setup_host
+
+    test 'should prolong token for host image' do
+      Setting[:token_duration] = 60
+      @host.set_token
+      past = Time.now
+      assert @host.token.expires > past
+      travel_to(past + 2.hours)
+      assert @host.token_expired?
+
+      perform_full_host_generate
+
+      assert_equal @host.token_expired?, true
+      assert @host.token.expires > past
+    end
+  end
+
   test 'should render help' do
     get :help, session: set_session_user
     assert_response :success


app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -15,6 +15,10 @@ class ISOGenerator
     def self.generate_full_host(host, opts = {}, &block)
       raise Foreman::Exception.new(N_('Host is not in build mode, so the template cannot be rendered')) unless host.build?
 
+      # update build token
+      token = host.token
+      token.update(expires: token.expires + Setting[:token_duration].minutes)
+
```
lzap, line 21: This code will fail when you try to build a bootdisk on a host that has no token (e.g. host which is not in build mode). In that case `host.token` is `nil` which will lead to NilException.

Also, you cannot create a token just by creating it, it would create new token without orchestration.

I think what we want to do here is: 

* if the host is in build mode and it has expired token
* then create a new one

Please use host's methods set_token, token_expired? and build? - see for example in here: https://github.com/theforeman/foreman/blob/develop/app/models/concerns/hostext/token.rb

lzap (changes requested):
This will not work, also after you apply recommended changes please do the same for host disk as well. Both host and full host has tokens in them.

app/services/foreman_bootdisk/iso_generator.rb
ares: Few questions, why do we need orchestration for token creation? I think the necessary orchestration is rather driven by `set_build` on host. Note that the host needs to be in the build mode for user to be able to download the bootdisk. When they try to get the bootdisk for hosts which is not in build mode, they receive (now ugly) error message about the fact.

IMHO the correct behavior is to prolong the existing token whenever a bootdisk is generated or create a new token, while keeping the check for build mode being enabled (meaning all provisioning services are prepare for provisioning)

domitea: What I understand this bug, the problem is with hosts that are prepared for building because they are missed the chance to provision due to token expiration. I know that prolonging token expiration is really bad security issue like in resetting password etc. However we can't do that because token is written into image so prolonging token is one logic solution for that. 

And as @ares I think that orchestration is not necessary. At least, this issue address to the hosts that has prepared orchestration (because build mode).

lzap: Right, prolonging is also an option. In that case, just update the code so it won't break inc ase token is `nil` and we are good to go.

lzap:
Please write a test case for this. Thanks.

domitea:
@lzap, I've updated code as you requested. The are almost greem, only one `connection refused` error. So I think it's ready for merge.

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -81,5 +86,13 @@ def error_rendering(exception)
       error("#{msg}: #{exception.message}")
       ::Foreman::Logging.exception(msg, exception)
     end
+
+    def prolong_token(host)
+      return unless host.token
```
ares, line 91: @lzap please keep me honest, I think host can be in build mode but token is nil, because it has expired and afterwards cleaned up (there's some cleanup mechanism for expired tokens). Perhaps the check should be that host is in build mode, if it is, either prolong an existing token or create a new one?

I'm not sure about this, so @domitea please investigate the token expiration/cleanup behavior and share your findings here.

domitea: I've checked the code of Token (https://github.com/theforeman/foreman/blob/develop/app/models/concerns/hostext/token.rb) and I don't think that's a problem because when the host is in build mode, it has a token as far as I understand the code.

Further explanation:

There is a `expire_token` method at Token. This method is only used at `refresh_token_on_build` that is called every time before validation of token (https://github.com/theforeman/foreman/blob/develop/app/models/concerns/hostext/token.rb#L11). And when you at the code to the refresh one, you can see that when there is a creating a new token when the host become in build mode. This applies to the new hosts and even for the old one when they are changing mode. 

What do you think, @lzap and @ares ?

lzap: > there's some cleanup mechanism for expired tokens

There actually isn't. Tokens are only generated or deleted when a host enters/exits build mode. There is not expiration outside of that AFAIK, unattended controller checks for expiration time and if it is expired it is treated as non-existing that's all.

So I think the line here should be:

```suggestion
      return if !host.token_expired? || Setting[:token_duration] == 0
```

I would suggest not to prolong tokens on each download, this could have some security implications. Lets prolong it only when it expired.

lzap:
Other than that looks good.

app/controllers/foreman_bootdisk/disks_controller.rb
domitea: I think it's a good idea @lzap! 
However there is still some little probability that user will download image just before (let's say couple of minutes) token expiration. I know that a really edge case and the probability is very small but It's there.

It's enough solution for @ares?

domitea:
I've updated PR with @lzap comment's solution because I think good for use case that this PR tries to solve.

app/controllers/foreman_bootdisk/disks_controller.rb
lzap: On a second thought I think if an attacker have ability to download bootdisk iso, then there is no harm in prolonging it on every download. It can be little confusing tho so I agree with prolonging it on every download, let's put a sentence on the Help screen for host bootdisks: Each download will renew provisioning token expiration when in use. 

Therefore I am expecting something like this:

    return if Setting[:token_duration] == 0 || host.token.nil?

lzap:
One idea: https://projects.theforeman.org/issues/32400

lzap:
Damn there are conflicts, please resolve them.

app/controllers/foreman_bootdisk/disks_controller.rb
ares: I agree and would prefer to prolong on every download. Each newly downloaded bootdisk should have the full provisioning window.

app/controllers/foreman_bootdisk/disks_controller.rb
```diff
@@ -60,5 +65,19 @@ def help; end
...
+      error("#{msg}: #{exception.message}")
+      ::Foreman::Logging.exception(msg, exception)
+    end
+
+    def prolong_token(host)
+      return if Setting[:token_duration] == 0 || host.token.nil?
+
+      # update build token
+      token = host.token
+      token.update(expires: token.expires + Setting[:token_duration].minutes)
```
lzap, line 80: There is a bug, on every download it adds token duration to the expiration, so in keeps adding on every request. You need to set expiration in relative to the current time and not the last seen expiration.

lzap (changes requested):
Does not work as expected unfortunately :-)

lzap (approved):
Thank you.

PR #107, opened 2021-06-09
fixes #32400 - Adding explanation of disabled bootdisks

domitea:
In the settings of Bootdisk plugin, you can disable types of boodisks. There wasn't any explanation and simple way how to distinguish what's disabled. This PR adds asterisk after name of type at the bootdisk help page and adds also what does it mean at the same page.


ares:
Why I'm fine with allowing these through settings, we should also make sure to adjust the texts for downstream as some types are not supported anymore.

lzap (approved):
TY

PR #108, opened 2021-08-23
fixes #31673 - Fixing error handling when host is not in build mode

domitea:
With current implementation, it's possible to download a bootdisk full host even for hosts that are not in build mode. So this scenario ends with the Foreman error every time for hosts that are not in build mode. This PR handles this error much better. It adds some error messages, refuses serving an image for API requests and also disables button for download.

### DIFF
--- a/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
@@ -37,8 +37,13 @@ module ForemanBootdisk
           host = @disk
           if params[:full]
             return unless bootdisk_type_allowed?('full_host')
-            ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
-              send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
+            begin
+              ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
+                send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
+              end
+            rescue ::Foreman::Exception => e
+              raise e unless e.code == 'ERF42-2893'
+              render_error json: { error: _('Host is not in build mode')}, status: :method_not_allowed
             end
           else
             return unless bootdisk_type_allowed?
--- a/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
+++ b/app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
@@ -82,8 +82,26 @@ module ForemanBootdisk
                                class: 'la'
                              )
 
+      full_host_blind_link = if bootdisk_authorized_for({controller: 'foreman_bootdisk/disks', action: 'full_host', id: host})
+                               link_to(_("Full host '%s' image") % host.name.split('.')[0],
+                                       '#',
+                                       class: 'la btn btn-info',
+                                       disabled: true,
+                                       title: _('Host is not in build mode')
+                               )
+                             else
+                               ''
+                             end
+
       actions << host_image_link if allowed_actions.include?('host')
-      actions << full_host_image_link if allowed_actions.include?('full_host')
+      if allowed_actions.include?('full_host')
+        if host.build?
+          actions << full_host_image_link
+        else
+          actions << full_host_blind_link
+        end
+      end
+
 
       actions << divider if (host_image_link.present? && allowed_actions.include?('host')) || (full_host_image_link.present? && allowed_actions.include?('full_host'))
 
--- a/app/controllers/foreman_bootdisk/disks_controller.rb
+++ b/app/controllers/foreman_bootdisk/disks_controller.rb
@@ -54,9 +54,14 @@ module ForemanBootdisk
       host = @disk
 
       prolong_token(host)
-
-      ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
-        send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
+      begin
+        ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
+          send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
+        end
+      rescue ::Foreman::Exception => e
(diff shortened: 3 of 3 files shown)


app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -37,8 +37,13 @@ def host
           host = @disk
           if params[:full]
             return unless bootdisk_type_allowed?('full_host')
-            ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
-              send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
+            begin
+              ForemanBootdisk::ISOGenerator.generate_full_host(host) do |iso|
+                send_file(iso, filename: "#{host.name}#{ForemanBootdisk::ISOGenerator.token_expiry(host)}.iso")
+              end
+            rescue ::Foreman::Exception => e
```
ares, line 44: why do we let it try to generate the bootdisk and send it if we can check the build mode of a host beforehand? the same below

lzap: FYI build mode is only required for full host I think (please test this but I believe this is correct).

Anyway, I do not agree with @ares here. What @domitea implemented is a typical Java-approach and it is a reasonable solution - the exception is thrown on the first line of generate_full_host.

app/helpers/concerns/foreman_bootdisk/hosts_helper_ext.rb
```diff
@@ -82,8 +82,26 @@ def host_action_buttons(host)
...
       actions << host_image_link if allowed_actions.include?('host')
-      actions << full_host_image_link if allowed_actions.include?('full_host')
+      if allowed_actions.include?('full_host')
+        if host.build?
+          actions << full_host_image_link
+        else
+          actions << full_host_blind_link
+        end
+      end
+
```
lzap, line 104: Nicely done, hiding it would cause confusion to existing users.

lzap (approved):
I am good with this, let me test it.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
```diff
@@ -37,8 +37,13 @@ def host
...
+              raise e unless e.code == 'ERF42-2893'
+              render json: { error: _('Host is not in build mode')}, status: :method_not_allowed
```
lzap, line 46: Please use `render_error` from the base controller, this does not create a valid JSON and CLI returns just:

```
[lzap@nuc cli]$ hammer bootdisk host --host allan-ocampo.dummy.lan --full true --file /dev/null --force
Failed to download host disk image:
  Error: 405 Method Not Allowed
```

With debug I can see JSON is returned but CLI ignores it, there is I think root element missing or something.

```
[ INFO 2021-09-03T10:05:10 API] GET /bootdisk/api/hosts/16
[DEBUG 2021-09-03T10:05:10 API] Params: {
    "full" => true
}
[DEBUG 2021-09-03T10:05:10 API] Headers: {
    :params => {
        "full" => true
    }
}
[DEBUG 2021-09-03T10:05:10 API] Using authenticator: HammerCLIForeman::Api::SessionAuthenticatorWrapper
[ERROR 2021-09-03T10:05:11 API] 405 Method Not Allowed
[DEBUG 2021-09-03T10:05:11 API] {
    "error" => "Host is not in build mode"
}
[DEBUG 2021-09-03T10:05:11 Exception] Using exception handler HammerCLIForeman::ExceptionHandler#handle_general_exception
[ERROR 2021-09-03T10:05:11 Exception] Error: 405 Method Not Allowed
Failed to download host disk image:
  Error: 405 Method Not Allowed
```

lzap (changes requested):
Other than that, works great.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
lzap: Granted that ideally there should be an exception class and the code should catch only specific class. Feel free to implement that if you want.

ares: Exceptions are slower thanks to the backtrace and should be generally used for, well, exceptional behavior. This PR updates the UI in a way it does not display the link to generate the bootdisk, so we intentionally disable it. Therefore it's not exception, it's something we know about and can easily verify before it fails (aka validation). The same thing `bootdisk_type_allowed?` check above.

I don't undestand why we'd prefer to use exception here comparing to normal program flow. This way we even need to be checking exception message and reraising it in case it's unknown code.

lzap: That's correct, but the meaning of this is that you should not use exception for state in hotspots, like loops. This case, you would not even measure the difference. We could probably argue for some time, but I am gonna stop now - I am fine if you insist on an if statement. I do not see anything wrong with exception approach. Truth is, `host` is available in this context so why not to use it.

domitea: I "copied" this situation from similar problem (https://github.com/theforeman/foreman_bootdisk/blob/a1d90df01320157ce5e95021ae9a633c0141006d/app/controllers/foreman_bootdisk/disks_controller.rb#L23) and I think it to use it there. Reason is simpler maintenance. The code that covers that host is not in build is already there (in `IsoGenerator`) so I made the better exception handling to say what's wrong for user. So in other words, I would like to leave as it is because I don't think it's worth make own Exception for one specific use case at one possibility. 

Also, when the host is not in built mode, I am able to make a `host image`. It makes sense because this type of image is not bonded to specific host.

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
domitea: Call has been updated

app/controllers/foreman_bootdisk/api/v2/disks_controller.rb
ares: FTR I don't mind, I'm not the maintainer of this plugin :-)

lzap: I don't mind either, I will accept whatever you just pushed Dominik. I am not maintainer of this plugin as well :-)

lzap (approved):
It works fine including CLI, thanks.

PR #110, opened 2021-11-09
Fix migration: Invalid single-table inheritance

m-bucher:
Given that the `FakeConfigTemplate` class was introduced for foreman<1.9, I believe it is save to remove it now.

Partly reverts 1e744832ea6f1f3eb19a2f2cfdad586a4bdf013f

PR #112, opened 2022-01-10
Fixes #34238 - resolve migration issues after core migration cleanup

lzap:
This reverts commit 8145219c02a0cad97a15aee727e56f3e8867c76a.

https://community.theforeman.org/t/luna-nightly-rpm-pipeline-322-failed/26760

### DIFF
--- a/db/migrate/20140522185700_change_templatekind_to_bootdisk.rb
+++ b/db/migrate/20140522185700_change_templatekind_to_bootdisk.rb
@@ -1,14 +1,22 @@
 # frozen_string_literal: true
 
 class ChangeTemplatekindToBootdisk < ActiveRecord::Migration[4.2]
+  class FakeConfigTemplate < ApplicationRecord
+    self.table_name = if ActiveRecord::Base.connection.migration_context.get_all_versions.include?(20_150_514_072_626)
+                        'templates'
+                      else
+                        'config_templates'
+                      end
+  end
+
   def self.up
     kind = TemplateKind.where(name: 'Bootdisk').first_or_create
 
     tmpl_h = Setting.find_by(name: 'bootdisk_host_template').try(:value)
     tmpl_g = Setting.find_by(name: 'bootdisk_generic_host_template').try(:value)
 
-    (Template.unscoped.where('name LIKE ?', '%Boot disk%') |
-      Template.unscoped.where(name: [tmpl_h, tmpl_g].compact)).each do |tmpl|
+    (FakeConfigTemplate.unscoped.where('name LIKE ?', '%Boot disk%') |
+      FakeConfigTemplate.unscoped.where(name: [tmpl_h, tmpl_g].compact)).each do |tmpl|
       tmpl.update_attribute(:template_kind_id, kind.id)
     end
   end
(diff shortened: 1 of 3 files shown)


db/migrate/20140522185700_change_templatekind_to_bootdisk.rb
```diff
@@ -1,14 +1,22 @@
 # frozen_string_literal: true
 
 class ChangeTemplatekindToBootdisk < ActiveRecord::Migration[4.2]
+  class FakeConfigTemplate < ApplicationRecord
+    self.table_name = if ActiveRecord::Base.connection.migration_context.get_all_versions.include?(20_150_514_072_626)
+                        'templates'
+                      else
+                        'config_templates'
+                      end
```
evgeni, line 9: I think this is wrong.

https://github.com/theforeman/foreman/commit/9851420902165fc23d0ea1f6d53922a5f6a37c6a dropped the old `20150514114044_migrate_ptables_to_templates.rb` migration, so the `include?` will be `false`, and you'd end up with `config_templates` name, not `templaes` for the table name.

evgeni: Hmm, no, it dropped only parts of that migration. So it actually could work.

@m-bucher could you have a look, was the removal in #110 purely cleanup, or was there a bug?

m-bucher: IIRC we had problems during migration in certain cases. Though I am not sure in which scenarios specifically :thinking:

lzap: Hmm but the statement queries `db/migrate/20150514072626_add_type_to_config_template.rb` and not `20150514114044_migrate_ptables_to_templates.rb`. The former migration is still present even today in develop.

lzap: Yeah, I did not notice the conversation after the initial comment...

Anyway, isn't safer to simply test for a class constant presence? Something like `if defined? ConfigTemplates` ?

lzap: I pushed another commit that feels safer to me, what you guys think? That does not depend on any migration numbers or whatever, it simply tests if older class constant exists, if it does not it is named the new way.

evgeni: > Hmm but the statement queries `db/migrate/20150514072626_add_type_to_config_template.rb` and not `20150514114044_migrate_ptables_to_templates.rb`. The former migration is still present even today in develop.

Heh. I failed on Ctrl+F and this silly notation of the migration version. You're right.

lzap:
Note: Squash on merge into "Fixes #34238 - resolve migration issues after core migration cleanup"

lzap:
So what do you guys say? Merge and new release to fix nightlies? Or other concerns?

lzap:
* https://github.com/theforeman/foreman-packaging/pull/7454
* https://github.com/theforeman/foreman-packaging/pull/7455

PR #113, opened 2022-01-19
Fixes #34287 - do not use STI in migrations

ezr-ondrej:
This disables STI for the Fake class.
Also we use SQL queries directly to not initialize the ActiveRecords


lzap:
The user reports it works: https://community.theforeman.org/t/error-installing-bootdisk/26913/12

I am going ahead and ignore "npm ERR! Maximum call stack size exceeded" because this needs to go into 3.1.

Many thanks for help!

PR #114, opened 2022-03-01
Fixes #34536 - delete temp directories recursively

lzap:
The deletion did not work as temp directories must be removed recursively. Also improved the time calculation with more readable Rails helpers.


adamruzicka (approved):
Makes sense, ACK

PR #116, opened 2022-03-17
Fixes #34637 - drop unattended setting

lzap:
SSIA


ekohl (approved):
Should this increase the minimum Foreman version to 3.3? Or will you make a note that `unattended` must be enabled?

lzap:
Good point, that is indeed 3.3 material. Rebased.

ekohl (approved):
Thanks!

PR #117, opened 2022-03-24
Fixes #34674 - deprecate non-DSL settings

lzap:
SSIA

Do not merge, does not work yet.

### DIFF
--- a/lib/foreman_bootdisk/engine.rb
+++ b/lib/foreman_bootdisk/engine.rb
@@ -17,15 +17,6 @@ module ForemanBootdisk
       app.routes_reloader.paths << "#{ForemanBootdisk::Engine.root}/config/routes/mount_engine.rb"
     end
 
-    initializer 'foreman_bootdisk.load_default_settings', before: :load_config_initializers do |_app|
-      table_exists = begin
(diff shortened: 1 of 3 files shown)


lib/foreman_bootdisk/engine.rb
```diff
@@ -65,6 +55,81 @@ class Engine < ::Rails::Engine
...
+
+            setting "bootdisk_isolinux_dir",
+              type: :string,
+              default: isolinux,
+              full_name: N_("ISOLINUX directory"),
+              description: N_("Path to directory containing isolinux images")
+
+            setting "bootdisk_syslinux_dir",
+              type: :string,
+              default: isolinux,
```
ezr-ondrej, line 83: I think Rubocop complains because of this oversight? :)
```suggestion
              default: syslinux,
```

ezr-ondrej:
Apart of the rubocop it seems to work and ready to merge to me :)

lib/foreman_bootdisk/engine.rb
lzap: That bastard actually useful? Hmmm.

PR #121, opened 2022-05-24
Version 20.0

Ron-Lavi:
Since require_foreman was changed earlier,
we actually need to bump this plugin's major version.


Ron-Lavi:
[test foreman_bootdisk]

PR #122, opened 2022-06-07
Test refactorings

stejskalleos:
- Use Github actions instead of Jenkins
- Fix 6.1 routing issues in tests

Jenkins removal PR: https://github.com/theforeman/jenkins-jobs/pull/225

PR #123, opened 2022-06-13
Use github actions instead of Jenkins

stejskalleos:


stejskalleos:
Will be done in https://github.com/theforeman/foreman_bootdisk/pull/122

PR #126, opened 2022-07-20
Fixes #35127 - migrate bootdisk templates to Foreman

nofaralfasi:
Related to this [PR](https://github.com/theforeman/foreman/pull/9314) on Foreman.
This PR applies a more complex solution for the missing description issue (compared to this [PR](https://github.com/theforeman/foreman_bootdisk/pull/128)), and it is intended for Satellite 6.13.


nofaralfasi:
Working on fixing the tests.

stejskalleos (changes requested):
* We should also remove `template_labels 'Bootdisk' => N_('Boot disk embedded template')` from `engine.rb`
* For failing tests you have to remove `setup_templates` in `test_helper` and then run tests again. Probably fix them same way as in discovery

nofaralfasi:
> * We should also remove `template_labels 'Bootdisk' => N_('Boot disk embedded template')` from `engine.rb`
> * For failing tests you have to remove `setup_templates` in `test_helper` and then run tests again. Probably fix them same way as in discovery

Thanks @stejskalleos 
There are also migrations related to the templates, for example - db/migrate/20171009225200_remove_duplicate_bootdisk_templates.rb.
It looks like it shouldn't be affected and we should keep it the way it is, but I wanted to make sure.

stejskalleos:
Blocked by https://github.com/theforeman/foreman_ansible/pull/548

test/test_plugin_helper.rb
```diff
@@ -29,7 +29,7 @@ def setup_routes
   end
 
   def setup_templates
-    load File.join(File.dirname(__FILE__), '..', 'db', 'seeds.d', '50-bootdisk_templates.rb')
+    load File.join(File.dirname(__FILE__), '..', '..', 'foreman', 'db', 'seeds.d', '070-provisioning_templates.rb')
```
stejskalleos, line 32: This wont work, as you can see from the failing tests:
```
Error:
ForemanBootdisk::Api::V2::DisksControllerTest::default API version 2#test_0002_path - /api/generic routes to #generic:
LoadError: cannot load such file -- /home/runner/work/foreman_bootdisk/foreman_bootdisk/foreman_bootdisk/test/../../foreman/db/seeds.d/070-provisioning_templates.rb
```

That file must be in this repository

stejskalleos (changes requested):

test/test_plugin_helper.rb
nofaralfasi: Yes yes, I know. I'm still trying to find a solution for this. This file cannot be in this repository since I deleted the templates and moved them to foreman. So I need to find another way, like using fixtures like I did in foreman_discovery.

stejskalleos: Well it can be still in this repository, but as a text fixture, in `test/static_fixtures/template.rb`

stejskalleos (approved):
:green_apple: **LGTM**

stejskalleos:
Thanks @nofaralfasi !

PR #128, opened 2022-07-27
Fixes #35282 - populate templates description from metadata

nofaralfasi:
This PR applies a simpler solution for the missing description issue(compared to this [PR](https://github.com/theforeman/foreman_bootdisk/pull/126)), and it is intended for Satellite 6.12.


Ron-Lavi (approved):
Thanks @nofaralfasi ! works well 👍🏼

PR #134, opened 2022-11-02
19.0.7 - Settings DSL

stejskalleos:


evgeni (approved):

evgeni:
This will also need a release in the 20.x series

evgeni:
> This will also need a release in the 20.x series

scratch that, 19 and 21 is sufficient

PR #136, opened 2022-12-18
Fixes #35794 - change bootdisk_allowed_types type to array

Ron-Lavi:
The setting value shows up as "subnetfull_host" whereas if we edit the value, we will be able to see that it is set as "subnet,full_host"

### DIFF
--- a/lib/foreman_bootdisk/engine.rb
+++ b/lib/foreman_bootdisk/engine.rb
@@ -130,7 +130,7 @@ module ForemanBootdisk
               description: N_("Installation media files will be cached for full host images")
 
             setting "bootdisk_allowed_types",
-              type: :string,
+              type: :array,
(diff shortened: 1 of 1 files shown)


nofaralfasi:
LGTM!
Should we move the Redmine issue to the foreman_bootdisk plugin?

Ron-Lavi:
Good catch @nofaralfasi ! I just changed the Redmine to bootdisk
@stejskalleos any other thoughts?
Once it's merged I will open the cherry-pick to `19.0-stable`

stejskalleos:
I would like to test it first, because I'm not sure if it is safe change, changing something from string to array could have big impact. I'll let you know

stejskalleos (approved):
:green_apple:  **LGTM**, works fine

stejskalleos:
Thanks @Ron-Lavi !

PR #137, opened 2022-12-21
Fixes #35794 - change bootdisk_allowed_types type to array

Ron-Lavi:
The setting value shows up as "subnetfull_host" whereas if we edit the value, we will be able to see that it is set as "subnet,full_host"

(cherry picked from commit 7c75d4da007628d840cac051f8601c9e395fc838)


stejskalleos (approved):

stejskalleos:
Thanks @Ron-Lavi

PR #139, opened 2022-12-22
Fixes #35888 - Full host image fails with bad request

stejskalleos:
Enable https connection when uri scheme is https

### DIFF
--- a/app/services/foreman_bootdisk/iso_generator.rb
+++ b/app/services/foreman_bootdisk/iso_generator.rb
@@ -217,7 +217,7 @@ module ForemanBootdisk
           else
             http_object = Net::HTTP
           end
-          http_object.start(uri.host, uri.port) do |http|
+          http_object.start(uri.host, uri.port, use_ssl: uri.scheme == 'https') do |http|
             request = Net::HTTP::Get.new(uri.request_uri, 'Accept-Encoding' => 'plain')
 
             http.request(request) do |response|


app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -217,25 +217,22 @@ def self.fetch(path, uri, limit = 10)
...
-                raise("Too many HTTP redirects when downloading #{uri}") if limit <= 0
-
-                fetch(path, response['location'], limit - 1)
-                # prevent multiple writes to the cache
-                write_cache = false
-              else
-                raise ::Foreman::Exception, N_(format("Unable to download boot file %{uri}, HTTP return code %{code}", uri: uri, code: response.code))
-              end
+
+          response = http_object.get_response(uri)
```
ares, line 221: the `Accept-Encoding` was added for this reason https://projects.theforeman.org/issues/25992, does it now work even without?

app/services/foreman_bootdisk/iso_generator.rb
```diff
@@ -217,25 +217,22 @@ def self.fetch(path, uri, limit = 10)
...
+
+          case response
+          when Net::HTTPSuccess then
+            response.body do |chunk|
```
ares, line 225: does body actually accept a block? the docs says differently, internally it calls read_body without passing a block or any argument

ares (changes requested):
I'm not necessarily requesting changes, just answers :-)

stejskalleos:
Found the issue, problem was in missing `use_ssl=true` for `uri` with `https` scheme

ares (approved):
Thanks, fixes the issue. Merging!

PR #147, opened 2023-06-27
Fixes #35888 - Full host image fails with bad request (#139)

stejskalleos:
Enable https connection when uri scheme is https

(cherry picked from commit 7ff3112dd29198408961618ab8604bc01c912892)


stejskalleos:
@nofaralfasi review please

PR #148, opened 2023-09-18
i18n - pulling from tx

stejskalleos:


locale/fr/foreman_bootdisk.po
```diff
@@ -256,7 +256,7 @@ msgid "Unable to generate disk %{kind} template: %{error}"
 msgstr "Impossible de générer le modèle %{kind} de disque : %{error}"
 
 msgid "Unable to generate disk template, %{kind} template not found."
-msgstr "Impossible de générer le modèle de disque, le modèle %%{kind} est introuvable."
+msgstr "Impossible de générer le modèle de disque, le modèle%{kind} est introuvable."
```
ekohl, line 259: Looks like this was intended
```suggestion
msgstr "Impossible de générer le modèle de disque, le modèle %{kind} est introuvable."
```

locale/ja/foreman_bootdisk.po
```diff
@@ -77,28 +78,28 @@ msgid "Download subnet generic image"
 msgstr "サブネットの汎用イメージのダウンロード"
 
 msgid "Ensure %{file} is readable (or update \"Grub2 directory\" setting)"
-msgstr "%{file} が読み取り可能であることを確認します (または「Grub2 ディレクトリー」設定を更新します)"
+msgstr "%%{file} が読み取り可能であることを確認します (または「Grub2 ディレクトリー」設定を更新します)"
```
ekohl, line 81: The `%%{TAG}` appears to be a repeating problem. I think I've also seen this in Transifex, so it isn't some pulling problem.

ekohl (commented):
There is no mechanism to push translations back from source to Transifex. The best way is to change it in Transifex, either via the UI or upload an updated .po file. Then pull again.

stejskalleos:
@ekohl updated the PR

app/assets/javascripts/foreman_bootdisk/locale/fr/foreman_bootdisk.js
```diff
@@ -109,7 +109,7 @@
         "L'hôte n'est pas en mode build, c'est pourquoi le modèle ne peut être rendu"
       ],
       "Unable to generate disk template, %{kind} template not found.": [
-        "Impossible de générer le modèle de disque, le modèle %%{kind} est introuvable."
+        "Impossible de générer le modèle de disque, le modèle%{kind} est introuvable."
```
ekohl, line 112: This line is correct in the .po file. I wonder why it's no longer in sync.

ekohl (changes requested):
As for Japanese, I think it's best to fix it up in Transifex. Let me know if you need help with that.

stejskalleos:
@ekohl I don't have permissions to edit strings in transifex, could you please add me to the project?

ekohl:
I can't quite figure that out right now (the permission screens are not intuitive), so I quickly fixed it up myself. I'd expect that if you pull again, it is correct.

stejskalleos:
@ekohl updated

ekohl:
I still see the same problems. In the past I've seen some issues where I needed to run the command multiple times, probably because it didn't register the timestamp changes properly.

stejskalleos:
>  I needed to run the command multiple times

Which one exactly? The one in the plugin or the two in the foreman?

ekohl:
Good question. I used https://github.com/ekohl/theforeman-plumbing/blob/master/bulk-locale-update and I just reran that whole script. Didn't dive deeper. I think my script is now outdated since it doesn't call the JS bits, so be warned.

ekohl:
Apologies for not paying attention to this for a while. When I now pull myself I don't see the `%%TAG` issue anymore so I think if you pull again, it should be correct.

stejskalleos:
let's continue in https://github.com/theforeman/foreman_bootdisk/pull/151

PR #150, opened 2023-11-28
Fixes ISO attach in vmware version 8 update 2

Manisha15:


chris1984:
@sbernhard that one is a dupe of this one which is now an RFE: https://bugzilla.redhat.com/show_bug.cgi?id=2247345

chris1984:
> This should fix https://bugzilla.redhat.com/show_bug.cgi?id=2247081 @chris1984

Will start testing

app/models/concerns/foreman_bootdisk/compute_resources/vmware.rb
```diff
@@ -40,7 +40,11 @@ def iso_attach(iso, vm_uuid)
           'connected' => true,
           'allow_guest_control' => true
         }
+	vm = find_vm_by_uuid(vm_uuid)
```
sbernhard, line 43: Can this be nil?

Should we have a block like
```
start_it_again = false
if vm.present? && vm.ready?
  vm.stop
  vm.reload
  start_it_again = true
end
client.vm_reconfig_cdrom options
vm.start if start_it_again
```

Manisha15: This cannot be nil otherwise all other definitions won't work as well like iso_upload, bootdisk_datastore, reconfig_cdrom etc.

chris1984 (approved):
Works great:

Before:
bootdisk_before

After:
bootdisk_after

@stejskalleos can this get a merge and a release so we can ship this upstream and downstream?

stejskalleos (approved):
:green_apple: **LGTM**

Didn't test it but code looks sane and @chris1984 confirmed functionality,
so let's get this in!

PR #165, opened 2024-08-19
i18n - pulling from tx

stejskalleos:


locale/fr/foreman_bootdisk.po
```diff
@@ -256,7 +256,7 @@ msgid "Unable to generate disk %{kind} template: %{error}"
 msgstr "Impossible de générer le modèle %{kind} de disque : %{error}"
 
 msgid "Unable to generate disk template, %{kind} template not found."
-msgstr "Impossible de générer le modèle de disque, le modèle %%{kind} est introuvable."
+msgstr "Impossible de générer le modèle de disque, le modèle%{kind} est introuvable."
```
ekohl, line 259: This looks like a regression. I've just applies this change to Transifex so if you pull again it should be good again
```suggestion
msgstr "Impossible de générer le modèle de disque, le modèle %{kind} est introuvable."
```

ekohl (changes requested):

stejskalleos:
Translations updated

ekohl (changes requested):
When updating translations, you need to make sure the `.mo` is also updated because that gets used at runtime.

stejskalleos:
> When updating translations, you need to make sure the .mo

I run precisely the same steps as I do with other plugins, but the `.mo` files are just not updated.

PR #167, opened 2024-09-09
Fixes #37792 - Safe mode and bootdisk_* template helpers

stejskalleos:
Using :extend_template_helpers instead of :allowed_template_helpers
fix the issue when rendering bootdisk_* template helpers in the Safe mode.

For steps to reproduce, see https://projects.theforeman.org/issues/37792


nofaralfasi (changes requested):
Failing tests are related.

stejskalleos:
Moving to draft, current changes breaks the full-host iso generation

stejskalleos:
@nofaralfasi ready for review

tested subnet, host image and full host images and all generated correctly

nofaralfasi:
This fix resolved the issue for me, but I'm unclear about why this change was necessary. @stejskalleos, could you please explain the reason for these changes? I’m assuming it was working previously, so what led to this problem?

stejskalleos:
@nofaralfasi I believe the reason is the recent move to Zeitwerk and its way of loading modules and classes.

With the original approach, the `ForemanBootdisk::Scope::Module` wasn't loaded before the Foreman's `render` method tried to access methods from it, causing the `no method` error.

nofaralfasi (approved):
I must admit that I still don’t fully understand some of these changes, such as the necessity of keeping the `ForemanBootdisk::Scope::Bootdisk` class. However, since this resolves the issue, I don't want to hold up the merge.

stejskalleos:
`Zeitwerk` follows a strict file structure-to-module/class mapping.

For example, having this code:
```ruby
module ForemanBootdisk
  module Templates
    module Scope
      # do something
    end
  end
end
```
It requires you to have the following files:
```
foreman_bootdisk/templates.rb
foreman_bootdisk/templates/scope.rb
```
Zeitwerk maps namespaces (modules or classes) to the corresponding file paths. When a constant like `ForemanBootdisk::Templates::Scope` is referenced for the first time, Zeitwerk looks for the file `templates/scope.rb` within the foreman_bootdisk directory. If the modules are defined within the same file, Zeitwerk won't be able to autoload `ForemanBootdisk::Templates::Scope` correctly.

cc @ofedoren please feel free to correct me if I'm wrong.

PR #169, opened 2024-10-06
refs #37792 - :template_name in correct scope

stejskalleos:
The template_name method must be under the Bootdisk scope; otherwise,
it overrides the method from Foreman and all templates (like Grub2 or Kickstart PXE)
have name ForemanBootdisk.

How to reproduce:
* Do not apply this patch
* Pull the latest bootdisk `master`
* Provision host (with kickstart PXE or Grub)
* See the rendered template


nofaralfasi (approved):
Works well for me.

PR #181, opened 2026-02-07
Add delete ISO method for vmware

sbernhard:
Requires: https://github.com/fog/fog-vsphere/pull/316


nadjaheitmann (approved):

stejskalleos (approved):
🍏 Code LGTM,
I haven't tested it personally, but we have ACK from @nadjaheitmann , so let's get this in.

PR #187, opened 2026-04-27
Fixes #39261 - Pin theforeman-rubocop in Gemfile only

ofedoren:


foreman_bootdisk.gemspec
```diff
@@ -27,6 +27,4 @@ Gem::Specification.new do |s|
               ['package.json']
 
   s.required_ruby_version = '>= 2.7', '< 4'
-
-  s.add_development_dependency 'theforeman-rubocop'
```
stejskalleos, line 31: I'm not against the change, but what's the reasoning here?
Is this causing an issue, or is it just a best practice?

ofedoren: Ehh, sorry I didn't write it here as well, but the idea is here:

https://github.com/theforeman/foreman_ansible/pull/789#issue-4335364319

adamruzicka (approved):
Ack

stejskalleos (approved):

PR #189, opened 2026-06-10
Require ruby 3.0 and higher

stejskalleos:


arvind4501 (approved):
I think It should also be released with this change

shubhamsg199 (approved):

.github/workflows/ruby_ci.yaml
```diff
@@ -12,7 +12,7 @@ concurrency:
 
 jobs:
   rubocop:
-    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v1
```
evgeni, line 15: We should add dependabot for that

evgeni (approved):

.github/workflows/ruby_ci.yaml
stejskalleos: * https://github.com/theforeman/foreman_bootdisk/pull/190
* https://github.com/theforeman/foreman_discovery/pull/682
* https://github.com/theforeman/foreman_google/pull/92
* https://github.com/theforeman/foreman_kubevirt/pull/199

PR #194, opened 2026-08-05
Fixes #39589 - Default bootdisk_mkiso_command to xorrisofs instead of genisoimage

ogajduse:
## Summary

CentOS Stream 10 dropped the genisoimage/cdrkit package entirely. `bootdisk_mkiso_command` still defaults to `"genisoimage"`, so ISO generation fails at runtime on a fresh EL10 install (nothing provides `/usr/bin/genisoimage` there, and it isn't installable).

Root cause dig: `mkisofs`, `genisoimage`, and `xorrisofs` aren't three different tools with varying compatibility -- on modern distros they're all the same `xorriso` binary, dispatching to the identical argument-emulation code path based on `argv[0]` (`xorriso/base_obj.c` in libisoburn: all four leafnames set `argument_emulation = 1` identically). Every flag `iso_generator.rb` passes (`-iso-level`, `-b`, `-c`, `-no-emul-boot`, `-boot-load-size`, `-boot-info-table`, `-eltorito-alt-boot`, `-e`) is handled by that same emulation layer. Confirmed by running the exact generated command against a CentOS Stream 10 container with `xorrisofs`.

`isohybrid` (used unconditionally right after ISO generation) is unaffected by this change -- still shipped via `syslinux` on EL10.

Filed as https://projects.theforeman.org/issues/39589 with the full investigation.

## Changes

- `lib/foreman_bootdisk/engine.rb`: default `bootdisk_mkiso_command` from `"genisoimage"` to `"xorrisofs"`, update setting description
- `app/services/foreman_bootdisk/iso_generator.rb`: update file-header comment listing runtime deps
- `test/unit/iso_generator_test.rb`: update hardcoded command-regexp expectation from genisoimage to xorrisofs
- `README.md`: update the two spots describing the ISO-tool dependency

## Test plan

- [x] Ran `test/unit/iso_generator_test.rb` against a Foreman `develop` core checkout (Postgres 17, schema loaded + seeded): 5 tests, 6 assertions, 0 failures, 0 errors

Fixes #39589


README.md
```diff
@@ -328,7 +328,7 @@ the Foreman UI.
 * _iPXE directory_ (`bootdisk_ipxe_dir`) points to the directory containing
   ipxe.lkrn
 * _ISO generation command_ (`bootdisk_mkiso_command`) is the name of
-  genisoimage/mkisofs on your OS
+  xorrisofs, genisoimage, or mkisofs on your OS
```
stejskalleos: Could we document the change between 9 and 10 here? For future generations.

stejskalleos (commented):
🍏 **LGTM**

Just one small nit.

README.md
ogajduse: addressed

stejskalleos (approved):

stejskalleos:
Thanks @ogajduse !

lib/foreman_bootdisk/engine.rb
```diff
@@ -112,9 +112,9 @@ class Engine < ::Rails::Engine
 
               setting "bootdisk_mkiso_command",
                 type: :string,
-                default: "genisoimage",
+                default: "xorrisofs",
```
evgeni, line 115: What about users who have bootdisk already installed? Will those remain on `genisoimage`?

stejskalleos: No, the default value will be changed.

However, if they change the setting to `something-else`, the value will still be `something-else`.

PR #195, opened 2026-08-26
Replace SSH with Proxmox Upload API for bootdisk ISO uploads

EnigmaXV:
- Replaced the SSH based  bootdisk ISO upload with the Proxmox Upload API.
- Updated iso_upload and iso_attach to accept and use the ISO Upload Storage selected by the user.
- Updated iso_delete to search all ISO-capable storages instead of using the first available ISO storage, which can be problematic when multiple ISO-capable storages exist. This is necessary because the selected storage_id is no longer available when the host is reloaded after provisioning.
- Added unit tests to cover the changes.

This PR depends on => https://github.com/theforeman/foreman_fog_proxmox/pull/543

### DIFF
--- a/app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
+++ b/app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
@@ -9,31 +9,34 @@ module ForemanBootdisk
         super + [:bootdisk]
       end
 
-      def iso_upload(iso, vm_uuid)
+      def iso_upload(iso, vm_uuid, storage_id: nil)
         server = find_vm_by_uuid(vm_uuid)
-        server.ssh_options = { password: fog_credentials[:proxmox_password] }
-        server.ssh_ip_address = proxmox_host
-        server.username = client.credentials[:current_user].split('@').first
-        server.scp_upload(iso, '/var/lib/vz/template/iso/')
-        server.reload
-        storage = storages(server.node_id, 'iso')[0]
-        storage.volumes.any? { |v| v.volid.include? File.basename(iso) }
+        upload_iso(server.node_id, storage_id, iso)
       end
 
       def iso_delete(iso, vm_uuid)
         server = find_vm_by_uuid(vm_uuid)
+        filename = File.basename(iso)
 
-        # delete the iso file from proxmox server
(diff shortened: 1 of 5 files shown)


stejskalleos:
@sbernhard I don't have access to the Proxmox instance on my setup; could you review it?

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -9,33 +9,34 @@ def capabilities
...
         server = find_vm_by_uuid(vm_uuid)
-        storage = storages(server.node_id, 'iso')[0]
-        volume = storage.volumes.detect { |v| v.volid.include? File.basename(iso) }
-        disks = server.disks.map { |disk| disk.split(":")[0] }.join(";")
+        storage = storages(server.node_id, 'iso').find { |candidate| candidate.identity == storage_id }
+        volume = storage&.volumes&.detect { |v| v.volid.include? File.basename(iso) }
+        raise ::Foreman::Exception, format(_('Could not find ISO %{iso} on storage %{storage}'), iso: File.basename(iso), storage: storage_id) unless volume
+
         server.update({ ide2: "#{volume.volid},media=cdrom" })
-        server.update({ boot: "order=ide2;#{disks}" })
```
m-bucher, line 38: I guess we still need this.
While testing I had the problem that the new VM did not boot from ISO, because it was set last in boot-order.

app/models/concerns/foreman_bootdisk/compute_resources/proxmox.rb
```diff
@@ -9,31 +9,34 @@ def capabilities
...
+        volumes = storages(server.node_id, 'iso').filter_map do |storage|
+          storage.volumes.find { |volume| File.basename(volume.volid) == filename }
+        end
+
+        if volumes.many?
+          raise ::Foreman::Exception,
+                format(_('Found multiple ISO images named %{iso}'), iso: filename)
+        end
+
+        volumes.first&.destroy
```
sbernhard: If no volumes is found, this work crash

sbernhard (approved):
Tested and works. Thanks.

